PR #53, opened 2016-04-11
Added a autodetect feature…

spankenstein:
for aws metadata info. Region, AZ, secgroup, and Subnet.

verified on test kube.

### DIFF
--- a/main.go
+++ b/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"log"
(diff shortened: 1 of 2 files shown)


FestivalBobcats:
Looks good to me. Is this rebased with master?

Also, if you don't mind, please replace any `log.Print...` with `core.Log.Info..` or `core.Log.<whatever_level>`

PR #56, opened 2016-04-12
Moved meta discover into the core package.

spankenstein:


spankenstein:
Well... slack seems to have exploded. So just let me know if you have any issues.

PR #61, opened 2016-04-20
Add full CRUD /nodes API

FestivalBobcats:
Under the hood:
Refactor Task system and move from api/ to core/. Resources can now
return an *Action that can be performed either synchronously or
asynchronously.


FestivalBobcats:
Coverage will have to wait... unit tests will come very soon.

PR #64, opened 2016-04-25
Capacity service

FestivalBobcats:


FestivalBobcats:
@gopherstein this will require you to add `--enable-capacity-service` to the container command.

This will be released shortly with 0.5.3.

PR #91, opened 2016-06-16
Add PerInstance field to Port, Addresses to Instance

FestivalBobcats:
New:
- add PerInstance (bool) to Port, which allows each Instance of a
  Component to have its own ServiceSet (external/internal K8S services,
  and Entrypoints)
- add Addresses (readonly) field to Instance to display addresses for
  each per_instance Port

Breaking:
- rename ComponentAddresses to Addresses (does not affect API)


FestivalBobcats:
Once merged, this will be 0.7.0.

PR #98, opened 2016-08-11
fix website link in Readme

mpkh:
currently it points https://github.com/supergiant/supergiant/blob/master/supergiant.io

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ Supergiant is API-based stateful container orchestration disguised as a
 developer-friendly application platform. It is based on
 [Kubernetes](https://github.com/kubernetes/kubernetes).
 
-[supergiant.io](supergiant.io)
+[supergiant.io](https://supergiant.io)
 
 ### Concepts
 


FestivalBobcats:
Much appreciated @mpkh! I know this is only a tiny change, but would you mind signing our contributor agreement before I merge?

http://supergiant.github.io/docs/agreements/

mpkh:
@FestivalBobcats hello... why... I mean can't you simply have some code of conduct?

briansage:
Yeah, it sucks. It's what any company with investors has to do. When submitting changes to Kubernetes, they require the same process.

We have a [code of conduct](https://supergiant.github.io/docs/community/code-of-conduct.html) that protects people from jerks being jerks, but signing the agreement protects _all_ changes from Evil Corp coming along, saying, "that belongs to meeee." It keeps all the contents of Supergiant open-source forever, so we have to be consistent about it.

mpkh:
@briansage @FestivalBobcats I don't feel right now like reading this for this change, I'm even not native English speaker and it's not very common what you really want even from me there... I'm ready to share my contribution even under WTFPL however feel free to close and steal this change if you need my passport data to get it there "OFFICIALLY"

PR #118, opened 2016-09-12
Add first run localhost instructions

briansage:
In order to help onboard newcommers, this commit updates `README.md`
with instructions for installing and configuring dependencies to run
Supergiant locally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,13 +5,25 @@ SUPERGIANT: Easy container orchestration using Kubernetes
 
 <!-- Links -->
 
-[Kubernetes]: https://github.com/kubernetes/kubernetes
-[Supergiant Site]: https://supergiant.io/
-[Tutorial AWS]: https://supergiant.io/blog/how-to-install-supergiant-container-orchestration-engine-on-aws-ec2?utm_source=github
-[Tutorial MongoDB]: https://supergiant.io/blog/deploy-a-mongodb-replica-set-with-docker-and-supergiant?urm_source=github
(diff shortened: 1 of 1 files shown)


robmurtha:
I'm starting to work with these instructions and found references to godep which was removed in the last merge.

briansage:
Gah, I'm too slow! Good catch @robmurtha.

FestivalBobcats:
Looks good @briansage. I do propose a change, though.

Since I was the guy who implemented SSL support, I needed to be able to use it locally. Outside of that, there's not really a purpose for SSL on `localhost`, so I think the section on generating the SSL files gets way out of our Readme scope. 

You can see [here](https://github.com/supergiant/supergiant/blob/master/pkg/core/core.go#L219-L221) how *unless* all 3 of the SSL-related settings are provided, SSL will be disabled.

We may want to make a note about how our AMI installer will generate a self-signed cert based on the EC2 instance hostname ([see here](https://github.com/supergiant/supergiant/blob/master/sgboot#L2-L6)). Other than that, though, the SSL options are available for anyone who wants to configure their CA-signed certs on their own setup in production.

briansage:
OK @FestivalBobcats, changes incorporated.

See [the updated section on Development](https://github.com/supergiant/supergiant/blob/f98c71e119b9ed33aed1a69147690ab3c8f517f5/README.md#development) to see how it reads now.

FestivalBobcats:
Looks good. Feel free to merge when Travis finishes running tests.

PR #122, opened 2016-09-14
Remove config.json from git, move to .example file

devrsw:
Since each install requires adding config.json to a local gitignore, this PR moves the config file to a json.example file, adding one more step to the dev process.


FestivalBobcats:
Yeah, I'm good with this. I think the .gitignore line should be `config/config.json` (without the leading slash).

devrsw:
Updated with leading slash removed.

PR #123, opened 2016-09-15
Move UI dependencies locally (WIP)

devrsw:
This branch moves JS, CSS, and font files to be hosted locally, instead of via CDN. Also, this sets up a file and routing structure to put incoming image, CSS, and JS assets as we move towards 1.0.


FestivalBobcats:
I'm going to merge this into the local_assets branch, so we can continue from their off the main repo.

ui/css/main.css
```diff
@@ -0,0 +1,6 @@
+@font-face {
+    font-family: 'Doppio One';
+    font-style: normal;
+    font-weight: 400;
+    src: local('Doppio One'), local('Doppio One'), url(/fonts/Doppio-One.ttf) format('truetype');
```
briansage: Instead of `.ttf`, we should use `.woff`. Honestly, they're supposed to be internally the same and all of today's browsers support both, but `.woff` is supposed to be the compressed and future-proof one.

css/main.css
```diff
@@ -0,0 +1,6 @@
+@font-face {
+    font-family: 'Doppio One';
+    font-style: normal;
+    font-weight: 400;
+    src: local('Doppio One'), local('Doppio One'), url(../fonts/Doppio-One.ttf) format('truetype');
```
briansage: I left another comment in `/ui/css/main.css` about `.ttf` vs. `.woff`. `.woff` is preferred.

pkg/ui/router.go
```diff
@@ -34,6 +34,18 @@ func init() {
...
 
+	cssHandler := http.FileServer(http.Dir("./ui/css/"))
+	jsHandler := http.FileServer(http.Dir("./ui/js/"))
+	fontsHandler := http.FileServer(http.Dir("./ui/fonts/"))
+	imagesHandler := http.FileServer(http.Dir("./ui/images/"))
+
+	http.Handle("/ui/css/", http.StripPrefix("/ui/css/", cssHandler))
+	http.Handle("/ui/js/", http.StripPrefix("/ui/js/", jsHandler))
+	http.Handle("/ui/fonts/", http.StripPrefix("/ui/fonts/", fontsHandler))
+  http.Handle("/ui/images/", http.StripPrefix("/ui/images/", imagesHandler))
```
briansage: Watch out for tabs vs. spaces.

ui/js/npm.js
```diff
@@ -0,0 +1,13 @@
+// This file is autogenerated via the `commonjs` Grunt task. You can require() this file in a CommonJS environment.
```
briansage: If we're not using this, it doesn't need to be in the repo, does it? Does something else depend on it?

ui/js/bootstrap.js
```diff
@@ -0,0 +1,2377 @@
+/*!
+ * Bootstrap v3.3.7 (http://getbootstrap.com)
```
briansage: Are we using this? If not, may as well not include it. Only include the minified version.

ui/views/layout.html
```diff
@@ -7,19 +7,15 @@
...
+      <link rel="stylesheet" href="/css/bootstrap-theme.min.css" integrity="sha384-fLW2N01lMqjakBkx3l/M9EahuwpSfeNvV63J5ezn3uZzapT0u7EYsXMjQV+0En5r" crossorigin="anonymous">
       <!-- Latest compiled and minified JavaScript -->
-      <script src="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.6/js/bootstrap.min.js" integrity="sha384-0mSbJDEHialfmuBBQP6A4Qrprq5OVfW37PRR3j5ELqxss1yVqOtnepnHVP9aJ7xS" crossorigin="anonymous"></script>
-
-      <!-- Font -->
-      <link href='https://fonts.googleapis.com/css?family=Doppio+One' rel='stylesheet' type='text/css'>
-      <!-- <link href="https://fonts.googleapis.com/css?family=Inconsolata" rel="stylesheet"> -->
+      <script src="/js/bootstrap.min.js" integrity="sha384-0mSbJDEHialfmuBBQP6A4Qrprq5OVfW37PRR3j5ELqxss1yVqOtnepnHVP9aJ7xS" crossorigin="anonymous"></script>
 
       <style>
```
briansage: Definitely should move all of this to `.css` files in a future commit.

devrsw: This is moved to a `.css` file already - am I missing something?

PR #139, opened 2016-09-26
Add Bootstrap Sass files

devrsw:
This PR adds:
* Sass files for Bootstrap (currently unchanged from Bootstrap 3.3.7
* Simple README in /css to show how to manually compile Sass
* Adds the Sass cache to gitignore
* New main.css based on Bootstrap


FestivalBobcats:
Awesome. So, unfortunately this comes with a lot of notes and some needed
changes, since our asset system is a bit complicated and hand-rolled. But aside
from that (which is all covered thoroughly below), we're on the right track.

1. We should `rm ui/assets/vendor/css/bootstrap*`

2. Do we still need `ui/assets/vendor/fonts/*` (specifically the glyphicons)?

3. Do we need `_bootstrap-compass.scss`, `_bootstrap-mincer.scss`, or
   `_boostrap-sprockets.scss`? I could be wrong, but I think those just apply
   for other frameworks.

4. Since we use go-bindata to compile assets directly into the binary, we will
   need to separate `.scss` files (and any future preprocessor code) from these
   directories somehow. I suggest:

      a. make a `ui/scss` dir, and move all the `.scss` files there

      b. change the line in README.md that says
  `go-bindata -pkg ui -o pkg/ui/generated_assets.go ui/...` to
  `go-bindata -pkg ui -o pkg/ui/generated_assets.go ui/assets ui/views`


**Also, important note:** all static files must already must go through a
go-bindata compilation step (as mentioned above). That means `.scss` files go
through 2 steps. You will compile from `ui/scss/*.scss` to
`ui/assets/css/main.css`, and will then run the go-bindata command.

It's probably best just to use this command to start up your local server:

`go-bindata -pkg ui -o pkg/ui/generated_assets.go ui/assets ui/views && go run main.go --config-file config/config.json`

devrsw:
All good points. In response:

1. Old bootstrap files in `css/bootstrap` have been removed.
2. We don't need the glyphicons, so I've removed them. We may have some missing file errors, but it's not important. `ui/assets/vendor/fonts/*` is needed, because we locally store the Doppio font instead of pulling from a CDN.
3. Those have been removed.
4. Moved `ui/scss/` outside of the assets directory, and updated the README. I've reorganized the assets folder slightly as well, while updating paths in `layout.html`.

devrsw:
Ok, I'm read for a code review here @FestivalBobcats. Some changes:

- main.css has been broken up into .scss files
- sass compiled and then packaged with go-bin-data
- readme updated
- minor restructuring of the assets folder, and updated paths in views

PR #191, opened 2017-01-23
Gitignore bindata. Fixes #188. (#190)

### LINKED ISSUE #188: Gitignore bindata
To not run into git conflicts I think it would be better to gitignore the bindata directory. Therefore this compilation step has to be added to the build.

FestivalBobcats:


FestivalBobcats:
Looks good, thanks @maikelvl.

PR #196, opened 2017-02-01
Update to CoreOS update scheme.

### LINKED ISSUE #194: CoreOS reboot strategy 'best-effort' deprecation
Accidentally came across a deprecation warning of the reboot strategy:

```console
Reboot strategy "best-effort" is deprecated and will be removed in the future.
Please explicitly set the reboot strategy to one of [off reboot etcd-lock]
See https://coreos.com/os/docs/latest/update-strategies.html for details on configuring reboot strategies.
```

gopherstein:
Fixes #194
This change sets nodes to not auto-reboot after update.
This can be modified after build by the user.


maikelvl:
Here is already a coreos key in those yams files.

PR #214, opened 2017-06-02
Version in CLI struct was taking precidence over cli.App

digitalsparky:
version always reported 0.0.0 because you were setting the Version on the CLI struct, not the underlying cli.App.


jordanrinke:
lgtm

gopherstein:
lgtm

PR #215, opened 2017-06-13
refactor openstack provider

jordanrinke:
Support for gophercloud latest version, ad to support v3 auth service including domain-id/domain-nam - no testing done, and need to omit struct bits if not in yaml - WIP


jordanrinke:
going to test a ui squash and merge

PR #221, opened 2017-06-16
SG AWS Bug Fix Update

### LINKED ISSUE #165: All supergiant AMI's are unavailable in us-east-2 (Ohio).
We're using us-east-2 (Ohio) for testing stuff, and we wanted to try out supergiant, but the AMI is unavailable in that datacenter. Github might not be the place to file this, but I didn't know where else to ask about it. If this is intended, that's fine, but I didn't see anything documented about it.

### LINKED ISSUE #173: Supergiant bucket names should be unique to prevent name conflicts in AWS.
If a cluster is created in 2 AWS regions with the same name, the install will fail because the bucket name will be already taken. Buckets should have a unique identifier.

gopherstein:
Fixes #165
Fixes #173

This change updates the default Kubernetes version to 1.5.7 and
updates the s3 bucket auth to v4. The AWS api client has been updated
to support the latest regions. The packer version has been updated in
Travis to support the lates AWS regions. When released AMI files will now be created for all new regions. This change also includes various
saftey checks for VPC injection. NOTE: Exsisting VPC builds are still considered
beta, so use this feature at your own risk.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,8 +18,8 @@ go:
 go_import_path: github.com/supergiant/supergiant
 
 install:
-  - wget https://releases.hashicorp.com/packer/0.10.1/packer_0.10.1_linux_amd64.zip
-  - unzip packer_0.10.1_linux_amd64.zip
+  - wget https://releases.hashicorp.com/packer/1.0.0/packer_1.0.0_linux_amd64.zip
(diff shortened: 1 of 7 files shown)


jordanrinke:
lgtm but I will be honest, I only scanned over the bash changes. :cake:

PR #223, opened 2017-06-21
Just determining where my fork conflict is

jordanrinke:


jordanrinke:
Found it, accidentally deleted a vendor.json row

PR #224, opened 2017-06-21
Refactor/Update OpenStack provider

jordanrinke:
OpenStack provider has been updated to support Newton/Ocata, the newest release of Gophercloud and v3 auth.


gopherstein:
Looks good to me. We need to do a testing pass day soon.

jordanrinke:
The failing test, the variables are right just in reversed order - a test that needs to be fixed later but apparently is randomly flakey on order of assignment.

PR #228, opened 2017-06-22
Digital Ocean Update

gopherstein:
This update:
Sets kubernetes version 1.5.7 as default on DO.
Allows for multi-master config on DO.


jordanrinke:
lgtm :cake:

PR #230, opened 2017-07-07
Update for AWS and EFS Feature

gopherstein:
This update moves the AWS provider to the unified Kubernetes
based version templates. It also adds support for EFS.

To use EFS:

- You can simply specify the bool build option during build.
```
  {
    "aws_config": {
      "region": "us-east-1",
      "build_elastic_filesystem": true,
      "vpc_ip_range": "172.20.0.0/16"
    },
    "cloud_account_name": "aws",
    "master_node_size": "m4.large",
    "name": "test",
        "ssh_pub_key": "my_ssh_key",
    "node_sizes": [
      "m4.large",
      "m4.xlarge",
      "m4.2xlarge",
      "m4.4xlarge"
    ]
  }
```

- Or you can specify a EFS ID
```
  {
    "aws_config": {
      "region": "us-east-1",
      "elastic_filesystem_id": "my_efs_id",
      "vpc_ip_range": "172.20.0.0/16"
    },
    "cloud_account_name": "aws",
    "master_node_size": "m4.large",
    "name": "test",
        "ssh_pub_key": "my_ssh_key",
    "node_sizes": [
      "m4.large",
      "m4.xlarge",
      "m4.2xlarge",
      "m4.4xlarge"
    ]
  }
```

In either case, SG will create a mount target on the share for any kubes it creates.
When your kube is deleted the target is removed, but the EFS share is not... This
must be done manually.


gopherstein:
I will need to fix some tests :-/

PR #232, opened 2017-07-09
Digital Ocean Node Name fix.

### LINKED ISSUE #231: all minion names are the same on DigitalOcean
I just discovered that all minion node names when using digital ocean are the same and follow a convention kube_name-minion . Is this by design? Kube doesn't seem to be recognizing it as multiple nodes, rather one node.

gopherstein:
Fixes #231

Node names are currently being created with the same name. This
updat ads a random string set at the end of all node names.


gopherstein:
Tested well... Looks good

PR #277, opened 2017-08-17
added all generated schemas except apps, fixed another kube sub issue

jordanrinke:


jordanrinke:
some troubleshooting stuff leaked in, removing it.

PR #353, opened 2017-09-13
Add new tracked event messages

alterx:
Apparently there are new messages for insufficient CPU and Memory, this commit allows the capacity service to react to these messages.


jordanrinke:
Thanks for the PR!, it looks like line 22 is missing a comma though.

alterx:
There you go 😄

gopherstein:
LGTM I will merge, release, and upstream to alpha as soon astroturfing that tests and release pass. 

Then you should be able to download the bin from the releases page.

PR #357, opened 2017-09-14
Supergiant Alpha Initial Push

gopherstein:
This push is the initial push for Supergiant Alpha, After this point,
master will be considered unstable, and releases will be created on
version branches.


briansage:
Because of these big changes, the /README.md or /docs/ should be updated with our change in forking and and PR policies.

Since the changes here are for an engaged community, it doesn't have to come in this PR, but we should add it soon.

jordanrinke:
lgtm

PR #362, opened 2017-09-18
Support custom tags and roles at the kube level and below.

jordanrinke:
Added the ability to specify an IAM role for master and node instances on AWS.

under the aws_config for kube you can now set the following:

aws_config {
  master_role: role_name,
  node_role: role_name
}

Also added the ability to set tags at the kube level that will tag any resources made for that kube.
It doesn't tag the S3 bucket (not sure if that is needed/wanted as it stores a single file)
This can be used with the following:

aws_config {
  tags: {tag_name: tag_value,
         another_tag: another_value}
}

No error checking is done on the tag for validity until it is applied, that is left to the user.


gopherstein:
LGTM

PR #419, opened 2018-02-06
Fixing DigitalOcean SSH Key Submission on "Create a Cluster"

### LINKED ISSUE #418: Creation of DigitalOcean cluster fails with “Error decoding JSON body: json: cannot unmarshal string into Go value of type []string”
I tested this on Chrome, using the debugging tools to find the error, thanks to another user. To reproduce this, make sure you have a working DO account with an API token and a public key uploaded. Create a DO cluster with any attributes, using the cloud account. The UI will spit out:
**Kube: DO Cluster
Error:Bad Request**

If you look in the developer console, you'll see that the response given for the network request is:
```
{
  "status": 400,
  "message": "Error decoding JSON body: json: cannot unmarshal string into Go struct field DOKubeConfig.ssh_key_fingerprint of type []string"
}
```

TheKLARKEN:
Fixes #418

This change adjusts the field in the schema used by the front end
from a string to an array, allowing the user to submit SSH key
fingerprints in a way that successfully adheres to the DigitalOcean
GoDo Go API client's expectations. Referenced in isse #418 by
@TheKLARKEN.

### DIFF
--- a/cmd/ui/assets/src/app/apps/apps-list/apps-list.component.html
+++ b/cmd/ui/assets/src/app/apps/apps-list/apps-list.component.html
@@ -12,7 +12,7 @@
 
       <div class="context-secondary-panel">
         <button type="button" class="btn btn-primary" [routerLink]="['/apps/new']">Deploy New App</button>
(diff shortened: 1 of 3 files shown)


jordanrinke:
lgtm

Issue #421, opened 2018-02-08
The scrollbar doesn't appear when creating a new cloud account


TheKLARKEN:
As brought to our attention by @manigandham, when creating a cloud account for a provider with a lengthy form, specifically seen when adding a GCE (Google Compute Engine) account, there's no scrollbar with which to reach the lower parts of the form:

![unnamed 1](https://user-images.githubusercontent.com/23109047/36003320-11652b06-0cf3-11e8-9512-59aa8c6b8b15.png)

This makes it necessary to zoom out on the page for large forms. I've been able to reproduce this issue and am tentatively planning to resolve it on Monday, ASAP.

PR #422, opened 2018-02-12
Added scrollbar to overflow content in cloud accounts pages.

TheKLARKEN:
Fixes issue #421. A simple css style was added to the app body to allow viewing of the extra content of cloud account pages, so that users can scroll down to fill out the rest of the form instead of zooming out, as referenced by @manigandham.


gopherstein:
LGTM

PR #428, opened 2018-02-18
Update config.json.example with new DO slugs

joelgriffith:
Adds in the new DigitalOcean slugs announced in January and removes the legacy ones as they're more expensive for the resources you get.


joelgriffith:
Thanks!

PR #446, opened 2018-03-13
[OpenStack] Security group support with OpenStack provider

### LINKED ISSUE #442: 1.0: Security groups aren't configurable for OpenStack kubes
As noted by @jrandall, there is currently no way for OpenStack users to have Supergiant configure security groups for their kube during provisioning. This means that if a user's default is to deny traffic, the kube won't be able to communicate. @jrandall would be able to give more insight on this and how/why it works this way.

Currently, we need to set up an OpenStack testing environment to work on this, which may take a few weeks to nail down.

Any feedback in the meantime as to how OpenStack users would like this to function would be very awesome!

colin-nolan:
Sets security groups on master and (minion) nodes.

Fixes #442.


gopherstein:
Rejecting Pull Please refer to Issue: https://github.com/supergiant/supergiant/issues/489

PR #474, opened 2018-05-08
UI graph and user change.

gopherstein:
This change removes negitive values before data
populates for the cluster charts. It also exposes
user API tokens in the user settings menu.


yegorch:
:+1:

PR #475, opened 2018-05-08
Added missing makefile commands:

yegorch:
1) clean - to quickly get rid of generated binary data
2) generate-bindata - to generate binary data wrapper.


gopherstein:
LGTM

PR #493, opened 2018-05-18
Fix helm release integration test

stgleb:
Sort config map keys to stabilize key traversal order and make test repeatable


gopherstein:
LGTM

PR #500, opened 2018-05-23
Wrap the db errors

saheienko:
Add to the database errors additional meta information.


yegorch:
@saheienko we don't really need this PR because whole current DB layer goes away.

saheienko:
@yegorch OK, closing it.

PR #502, opened 2018-05-23
Add JWT tokens for user authentication and authorization

stgleb:


pkg/user/domain.go
```diff
@@ -14,8 +15,8 @@ type User struct {
 
 type Repository interface {
 	GetAll(context.Context) ([]User, error)
-	Get(ctx context.Context, login string) (*User, error)
```
yegorch: It would be a good idea to keep named params here, because it gives a developer an idea what kind of string is expected here.

pkg/user/middleware.go
```diff
@@ -0,0 +1,23 @@
+package user
+
+import (
+	"net/http"
+
+	sgjwt "github.com/supergiant/supergiant/pkg/jwt"
+)
+
+const supergiantAuthHeader = "SGTOKEN"
```
yegorch: Shouldn't it be standard "Authentication" header now?

pkg/user/middleware.go
```diff
@@ -0,0 +1,37 @@
...
+
+import (
+	"net/http"
+
+	"strings"
+
+	sgjwt "github.com/supergiant/supergiant/pkg/jwt"
+)
+
+func AuthMiddleware(tokenService sgjwt.TokenService, next http.Handler) http.Handler {
```
yegorch: Middleware should not be present in package that provides business logic, please move that to the API package.

yegorch:
LGTM

PR #503, opened 2018-05-24
Use helm client insead of running helm pod

saheienko:
Add helm client, that helps to get rid of the "helm job pod". 
A structure for helm repositories was updated. Also, apps (releases) with all statuses are now displayed.

Used packages:
```
k8s.io/kubernetes/helm@v2.8.2
```


pkg/core/helm_releases.go
```diff
@@ -139,204 +154,59 @@ func (c *HelmReleases) Delete(id *int64, m *model.HelmRelease) ActionInterface {
...
+				if err != nil {
+					return errors.Wrap(err, "build helm client")
 				}
+
+				resp, err := hclient.DeleteRelease(m.Name, helm.DeletePurge(true))
+				if err != nil {
+					return errors.Wrapf(err, "delete %s release", m.Name)
+				}
+
+				c.Core.Log.Debugf("DEBUG - Delete %s release: %s", m.Name, resp.Info)
```
yegorch: @saheienko Could you please change logging statements to logrus?

pkg/helm/client.go
```diff
@@ -0,0 +1,114 @@
...
+)
+
+var (
+	tillerPodLabels = labels.Set{"app": "helm", "name": "tiller"}
+)
+
+// TODO: remove after refactoring
+// keep track of the created clients
+var (
+	clients map[string]*client
```
yegorch: @saheienko  Have you thought about use case  when that map is accessed/modified concurrently?

yegorch:
@saheienko Any tests?

pkg/core/helm_releases.go
saheienko: Updated

pkg/helm/client.go
saheienko: Updated, use sync.Map here.

saheienko:
@yegorch, integration tests will be added later.

PR #505, opened 2018-05-24
Add profiles for nodes and clusters

stgleb:
Extract common configurations to separate entities called profiles
that will enable to store them and provide to user e.g cpu intensive workload
or high memory capacity profile.


pkg/cloud_account/domain.go
```diff
@@ -13,3 +15,12 @@ type CloudAccount struct {
 	Provider    provider.Name
 	Credentials Credentials
 }
+
+//Repository is used to abstract domain storage from implementation
+type Repository interface {
```
yegorch: How come this interface is here? I've deleted it yesterday.

pkg/cloud_account/service.go
```diff
@@ -3,73 +3,28 @@ package cloud_account
...
-
 	"github.com/pkg/errors"
-	"github.com/sirupsen/logrus"
-	"github.com/supergiant/supergiant/pkg/storage"
 )
 
-// Service holds all business logic related to cloud accounts
 type Service struct {
-	Repository storage.Interface
+	Repository Repository
```
yegorch: Should use storage.Interface

pkg/cloud_account/storage_integration_test.go
```diff
@@ -0,0 +1,79 @@
+//+build integration
+
+package cloud_account
```
yegorch: Should be deleted already.

pkg/profile/kube_profile_service_test.go
```diff
@@ -0,0 +1,147 @@
+package profile
+
+import (
+	"context"
+	"testing"
+
+	"github.com/pkg/errors"
+)
+
+type fakeStorage struct {
```
yegorch: Why not using testify mocks?

pkg/profile/kube_profile.go
```diff
@@ -0,0 +1,10 @@
+package profile
+
+type KubeProfile struct {
```
saheienko: It is better to use a Kube, because exported names are qualified by their package names.

yegorch: @saheienko Usually it is. But we already have Kube struct so it would be source of confusion

saheienko: @yegorch Do you mean model.Kube?

pkg/profile/kube_profile_service_test.go
stgleb: Ok

pkg/profile/kube_profile.go
yegorch: @saheienko yes, model.Kube would be deprecated as everything in model package.

yegorch: @stgleb add validations using govalidator

test/integration/profile/kube_profile_test.go
```diff
@@ -0,0 +1,135 @@
...
+		t.Errorf("Unepexpected error while creating kube profile %v", err)
+	}
+
+	kube2, err := service.Get(ctx, kube.Id)
+
+	if err != nil {
+		t.Errorf("Unexpected error while getting kube profile %v", err)
+	}
+
+	if kube.Id != kube.Id || kube.KubernetesVersion != kube2.KubernetesVersion {
```
saheienko: kube.Id != kube.Id

test/integration/profile/node_profile_test.go
```diff
@@ -0,0 +1,116 @@
...
+		t.Errorf("Unepexpected error while creating node profile %v", err)
+	}
+
+	node2, err := service.Get(ctx, node.Id)
+
+	if err != nil {
+		t.Errorf("Unexpected error while getting node profile %v", err)
+	}
+
+	if node.Id != node.Id || node.Image != node2.Image || node.Size != node2.Size {
```
saheienko: node.Id != node.Id

stgleb:
@gopherstein Will be done

stgleb:
@TheKLARKEN I guess I've removed that in latest commits. If you found something else - please leave comments on the code.

PR #510, opened 2018-05-29
WIP Refactored storage to reduce boilerplate

yegorch:


pkg/storage/storage.go
```diff
@@ -5,48 +5,62 @@ import (
...
+	Put(ctx context.Context, prefix string, key string, value interface{}) error
 	Delete(ctx context.Context, prefix string, key string) error
 }
 
 type ETCDRepository struct {
 	cfg clientv3.Config
 }
 
-func (e *ETCDRepository) Get(ctx context.Context, prefix string, key string) ([]byte, error) {
+func (e *ETCDRepository) Get(ctx context.Context, prefix string, key string, resultType reflect.Type) (interface{}, error) {
```
saheienko: Accept reflect.Type as a parameter and return an interface, rly?

yegorch: What’s wrong with that? Also, how would you implement GetAll efficiently in that case?

saheienko: 1.
```
o := new(Obj)
// get object
i, err := s.Get(ctx, key, reflect.TypeOf(o))
if err != nil {
...
}

// cast an interface to object
o1, ok := i.(Obj)
if !ok {
...
}

// use object
```

Instead of
```
o := new(Obj)
// get object
err := s.Get(ctx, key, reflect.TypeOf(o))
if err != nil {
...
}

// use object
```
2. Do a GetAll method depends on a Get?

stgleb: @yegorch I think we have to change our view on that problem. You try to create a general interface that could cover all existing and future types in our project. I think we had better to be specific in interfaces and general in implementation. So each package should have an interface called storage inside of itself like

```
	GetAll(ctx context.Context, prefix string) ([]*Node, error)
	Get(ctx context.Context, prefix string, key string) (*Node, error)
	Put(ctx context.Context, prefix string, key string,*Node) error
	Delete(ctx context.Context, prefix string, key string) error
```

And we should have single package storage that will implement those methods for all possible types, so when you want to introduce the new package - you declare the interface for this type  and implement actual  methods inside storage type.I think this is more "Go way", so interfaces are lightweight, specific and are declared only where they are used.

yegorch:
Closing this PR

PR #513, opened 2018-05-30
Extract template initialization to separate package

stgleb:
Move all stuff related to creating actual template object to separate package
Make provider package has less dependencies


pkg/provider/template/template.go
```diff
@@ -0,0 +1,56 @@
...
+	"github.com/sirupsen/logrus"
+
+	"fmt"
+
+	"github.com/supergiant/supergiant/bindata"
+	"github.com/supergiant/supergiant/pkg/provider"
+)
+
+var (
+	Templates map[string]*template.Template
```
saheienko: It is a bad practice to use global variables ([magic is bad](https://dave.cheney.net/2017/06/11/go-without-package-scoped-variables)). Even if you in need, will be better to use a named type (or even a struct) and control access to it via methods.

pkg/provider/template/template.go
```diff
@@ -0,0 +1,56 @@
...
+)
+
+func init() {
```
saheienko: On my opinion, it is better to use a constructor here instead of the init func, and "inject" this templates to providers.

pkg/provider/template/template.go
```diff
@@ -0,0 +1,56 @@
...
+	kubeVersions := []provider.K8SVersion{provider.K8S15, provider.K8S16, provider.K8S17, provider.K8S18}
+	NodeRoles := []provider.NodeRole{provider.Master, provider.Minion}
```
saheienko: It is better to use provider.NodeRoles() func, that return all the roles. (what if we add an another role, will you remember to update it here?)
The same is for kube versions.

pkg/provider/template/template.go
```diff
@@ -0,0 +1,56 @@
...
+	Templates = make(map[string]*template.Template)
+
+	for _, nodeRole := range NodeRoles {
+		for _, kubeVersion := range kubeVersions {
+			mversion := strings.Split(string(kubeVersion), ".")
```
saheienko: Will it be better to use struct with methods Major(), Minor(), String() etc?

pkg/provider/aws/create_node.go
```diff
@@ -5,16 +5,18 @@ import (
...
-	"text/template"
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/ec2"
 
-	"github.com/supergiant/supergiant/bindata"
+	"github.com/Sirupsen/logrus"
+
+	"strings"
```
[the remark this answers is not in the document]
stgleb: Yeah, just missed that when run goimports

pkg/provider/aws/create_node.go
```diff
@@ -26,18 +28,12 @@ func (p *Provider) CreateNode(m *model.Node, action *core.Action) error {
...
-	template, err := template.New("minion_template").Parse(string(userdataTemplate))
-	if err != nil {
-		return err
-	}
+	minionFileName := fmt.Sprintf("config/providers/common/%s.%s/minion.yaml)", mversion[0], mversion[1])
+	tpl := template.Templates[minionFileName]
+
 	var userdata bytes.Buffer
-	if err = template.Execute(&userdata, m); err != nil {
+	if err := tpl.Execute(&userdata, m); err != nil {
```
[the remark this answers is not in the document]
stgleb: Ok, I'll handle the case more gracefully.

pkg/provider/constants.go
```diff
@@ -1,11 +1,23 @@
...
 	AWS          Name = "aws"
 	DigitalOcean Name = "digitalocean"
 	Packet       Name = "packet"
 	GCE          Name = "gce"
 	OpenStack    Name = "openstack"
+
+	K8S15 K8SVersion = "1.5"
+	K8S16 K8SVersion = "1.6"
+	K8S17 K8SVersion = "1.7"
+	K8S18 K8SVersion = "1.8"
```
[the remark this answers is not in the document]
stgleb: I am not sure about the way we should store versions. I've cut patch because is we have only major and minor in directory, so they are pretty redundant regarding getting right yml file.

pkg/provider/constants.go
```diff
@@ -1,11 +1,23 @@
 package provider
 
-type Name string
+type (
+	Name       string
```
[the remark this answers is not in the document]
stgleb: Ok, even though this is not my code originally

pkg/provider/template/template.go
stgleb: Yes, this is just WIP so far

yegorch:
@stgleb @gopherstein This PR isn't needed anymore as in nearest future we plan to get rid from templates in favor of SSH based provisioners.

PR #515, opened 2018-06-01
Add data layer for a helm repositories

saheienko:
Add handlers and a service, that uses etcd storage, for a helm repositories.


internal/helm/repository/endpoint.go
```diff
@@ -0,0 +1,97 @@
+package repository
```
yegorch: It should be part of helm package.

internal/helm/repository/endpoint.go
```diff
@@ -0,0 +1,97 @@
...
+	"github.com/supergiant/supergiant/pkg/model/helm"
+)
+
+// Controller is a http controller for a helm repositories.
+type Controller struct {
+	svc *Service
+}
+
+// New constructs a Controller for helm repositories.
+func New(svc *Service) (*Controller, error) {
```
yegorch: Please change to NewService and remove error in return

internal/helm/repository/endpoint.go
```diff
@@ -0,0 +1,97 @@
...
+
+	"github.com/gorilla/mux"
+	"github.com/sirupsen/logrus"
+	"gopkg.in/asaskevich/govalidator.v8"
+
+	"github.com/supergiant/supergiant/pkg/model/helm"
+)
+
+// Controller is a http controller for a helm repositories.
+type Controller struct {
```
yegorch: Please rename it to Endpoint so it would be uniform across packages.

internal/helm/repository/service.go
```diff
@@ -0,0 +1,85 @@
...
+)
+
+// Service manages helm repositories.
+type Service struct {
+	prefix  string
+	storage storage.Interface
+}
+
+// NewService constructs a Service for helm repository.
+func NewService(s storage.Interface) (*Service, error) {
```
yegorch: Again, please don't return error if error can't happen there

internal/helm/repository/endpoint.go
saheienko: Such king of stuff will be created for releases. Will we store them in one package?

internal/helm/repository/endpoint.go
saheienko:
```
// Endpoint is a http controller for account entity
```
Kind of strange naming... But, OK.

internal/helm/repository/endpoint.go
saheienko: I'll rename it. BTW, it would be better to keep a service and an endpoint separately.

About an error, I've supposed it is a stable pattern for a constructor: __New(opts) (Obj, error)__, because now you might not have an error, but WHEN you will add some opts and you need to return an error, you should updated/modify code where it used.
The reason to add an error now, that it isn't affect the complexity.

internal/helm/repository/handler.go
```diff
@@ -0,0 +1,91 @@
...
+		logrus.Errorf("handler: create %s helm repo: %v", repo.Name, err)
+		http.Error(w, "failed to create", http.StatusInternalServerError)
+		return
+	}
+}
+
+// Get retrieves a helm repository.
+func (h *Handler) Get(w http.ResponseWriter, r *http.Request) {
+	repoName := mux.Vars(r)["repoName"]
+
```
stgleb: looks like this print is redundant here

internal/helm/repository/handler.go
```diff
@@ -0,0 +1,91 @@
...
+	println("reponame:", repoName)
+
+	repo, err := h.svc.Get(r.Context(), repoName)
+	if err != nil {
```
stgleb: I think we should discuss something handleError utility function to write errors to writer and/or logger.

internal/helm/repository/handler.go
```diff
@@ -0,0 +1,91 @@
...
+	if !ok {
+		http.Error(w, "invalid repository", http.StatusBadRequest)
+		return
+	}
+
+	if err = h.svc.Create(r.Context(), repo); err != nil {
+		logrus.Errorf("handler: create %s helm repo: %v", repo.Name, err)
+		http.Error(w, "failed to create", http.StatusInternalServerError)
+		return
+	}
```
stgleb: Does helm repo created synchronously? If now StatusAccepted  should  be there.

internal/helm/repository/handler.go
```diff
@@ -0,0 +1,91 @@
...
+// Delete removes a helm repository from the storage.
+func (h *Handler) Delete(w http.ResponseWriter, r *http.Request) {
+	repoName := mux.Vars(r)["repoName"]
+
+	err := h.svc.Delete(r.Context(), repoName)
+	if err != nil {
+		logrus.Errorf("handler: delete %s helm repository: %v", repoName, err)
+		http.Error(w, "", http.StatusInternalServerError)
+		return
+	}
```
stgleb: The same question about nature of this action sync or async?

internal/helm/repository/handler.go
saheienko: Yes, we should.

internal/helm/repository/handler.go
saheienko: Yeap, it is.

PR #516, opened 2018-06-04
Add kube package

stgleb:


pkg/kube/endpoint.go
```diff
@@ -0,0 +1,83 @@
...
+	service *KubeService
+}
+
+func (e *KubeEndpoint) GetKube(w http.ResponseWriter, r *http.Request) {
+	vars := mux.Vars(r)
+	kubeId := vars["id"]
+
+	kube, err := e.service.Get(r.Context(), kubeId)
+
+	if err != nil {
```
yegorch: What if we have internal server error?

pkg/kube/endpoint.go
```diff
@@ -0,0 +1,83 @@
...
+	}
+
+	err = e.service.Create(r.Context(), kube)
+
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	w.WriteHeader(http.StatusCreated)
```
yegorch: Shouldn' it be accepted?

pkg/kube/endpoint_test.go
```diff
@@ -0,0 +1,30 @@
...
+	"net/http/httptest"
+	"testing"
+)
+
+func TestKubeProfileEndpointCreateProfile(t *testing.T) {
+	endpoint := &KubeEndpoint{}
+	kube := &Kube{
+		Name:              "",
+		KubernetesVersion: "1.8.7",
+		MasterPublicIP:    "12.34.56.78",
```
yegorch: Why a single public IP? It could have a lot of them.

pkg/kube/service.go
```diff
@@ -0,0 +1,76 @@
...
+import (
+	"context"
+	"encoding/json"
+
+	"github.com/supergiant/supergiant/pkg/storage"
+)
+
+type KubeService struct {
+	prefix      string
+	kubeStorage storage.Interface
```
yegorch: Just storage is enough.

pkg/kube/service.go
```diff
@@ -0,0 +1,76 @@
...
+
+import (
+	"context"
+	"encoding/json"
+
+	"github.com/supergiant/supergiant/pkg/storage"
+)
+
+type KubeService struct {
+	prefix      string
```
yegorch: Maybe the prefix could be /supergiant/kube ? We should revisit it...

pkg/storage/mock.go
```diff
@@ -0,0 +1,47 @@
+package storage
```
yegorch: We already have that mock, see testutils/mock.go

pkg/kube/endpoint.go
```diff
@@ -0,0 +1,83 @@
...
+
+import (
+	"encoding/json"
+	"net/http"
+
+	"github.com/gorilla/mux"
+	"gopkg.in/asaskevich/govalidator.v8"
+)
+
+type KubeEndpoint struct {
```
saheienko: Endpoint (or Handler) would be enough. 
Usage:
```
k kube.KubeService
vs
k kube.Service
```

pkg/kube/service.go
saheienko: Nope, "supergiant" wil be set as a rootPath in a storage.

pkg/kube/endpoint.go
stgleb: Ok, I agree

pkg/kube/endpoint.go
stgleb: @yegorch I need your error package for that))

yegorch:
LGTM

PR #519, opened 2018-06-08
Add node task

stgleb:


scripts/gce/install_certificates.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
```
yegorch: @stgleb why are all scripts in gce? Are they specific to GCE only? I doubt it.

scripts/gce/install_tiller.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
```
yegorch: Code duplication, in script above there is already tiler. It should be run as two seprate tasks in a chain.

scripts/one_shot.sh
```diff
@@ -0,0 +1,108 @@
+#!/bin/bash -e
```
yegorch: It looks like AWS specific script. Why is it in this folder? Also, why do we need it?

tasks/util.go
```diff
@@ -0,0 +1,22 @@
+package tasks
```
yegorch: Please rename as it would conflict with machinery's package tasks.

tasks/digitalocean/node.go
```diff
@@ -0,0 +1,138 @@
...
+
+	"bytes"
+	"context"
+	"github.com/supergiant/supergiant/pkg/runner"
+	"github.com/supergiant/supergiant/pkg/runner/command"
+	"github.com/supergiant/supergiant/pkg/runner/ssh"
+	"github.com/supergiant/supergiant/tasks"
+)
+
+type Task struct {
```
yegorch: Is it possible to rename it to something else? Task is an already overloaded name.

tasks/digitalocean/node.go
```diff
@@ -0,0 +1,138 @@
...
+
+func (t *Task) runTemplate(ctx context.Context, tpl *template.Template, cfg taskConfig) error {
+	buffer := new(bytes.Buffer)
+	err := t.kubeletScript.Execute(buffer, cfg)
+
+	if err != nil {
+		return err
+	}
+
+	for {
```
yegorch: Why do you do it in endless for loop?

stgleb: @yegorch We don't know count of commands in advance.

scripts/gce/install_certificates.sh
stgleb: @yegorch We have separate scripts for aws, digitalocean and openstack and separate folder for gce. That's why I've moved them to separate folder.

tasks/digitalocean/node.go
stgleb: @yegorch Ok, jobs can be better for that.

tasks/digitalocean/node.go
yegorch: It is a bad idea to execute line by line. One command == one runner call is more reliable way.

saheienko:
@stgleb Can you make a separate PR for changes of the runner package?

stgleb:
@saheienko That would be not easy to split it on this stage, but I agree that these changes should go in separate pull requests.

saheienko:
Is it OK?
```
?       github.com/supergiant/supergiant/jobs   [no test files]
?       github.com/supergiant/supergiant/jobs/digitalocean      [no test files]
```

saheienko:
Can you fix it?
```
pkg/jobs/digitalocean/node.go:9:2: cannot find package "github.com/supergiant/supergiant/pkg/jobsjobs" in any of:
       /home/unknown/go/src/github.com/supergiant/supergiant/vendor/github.com/supergiant/supergiant/pkg/jobsjobs (vendor tree)
       /home/unknown/goroot/go1.10.3/src/github.com/supergiant/supergiant/pkg/jobsjobs (from $GOROOT)
       /home/unknown/go/src/github.com/supergiant/supergiant/pkg/jobsjobs (from $GOPATH)
```

pkg/jobs/util.go
```diff
@@ -0,0 +1,22 @@
+package jobs
+
+import (
+	"io/ioutil"
+	"text/template"
+)
+
+func ReadTemplate(fileName, templateName string) (*template.Template, error) {
+	data, err := ioutil.ReadFile(fileName)
```
saheienko: Do you plan to distribute supergiant with a set of templates?

scripts/start_kube_proxy.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+echo "{{ .ConfigFile }}" > /etc/kubernetes/config.json
```
saheienko: For what reason you store a kubernetes config file as a separate template?
```
scripts/start_kube_proxy.sh
```
Is it a shell script? It looks like a template file.

scripts/templates/kubelet_service.conf.tpl
```diff
@@ -0,0 +1,39 @@
...
+        -v /var/run:/var/run:rw \
+        -v /var/lib/docker/:/var/lib/docker:rw \
+        -v /var/lib/kubelet/:/var/lib/kubelet:shared \
+        -v /var/log:/var/log:shared \
+        -v /srv/kubernetes:/srv/kubernetes:ro \
+        -v /etc/kubernetes:/etc/kubernetes:ro \
+        gcr.io/google-containers/hyperkube:v{{ .KubernetesVersion }} \
+        /hyperkube kubelet --allow-privileged=true \
+        --cluster-dns=10.3.0.10 \
+        --cluster_domain=cluster.local \
```
saheienko: Why do you hardcode a dns address and a domain name?

scripts/templates/config.json.tpl
```diff
@@ -0,0 +1,10 @@
+{
+  "bind-address": "{{ .MasterPrivateIP }}",
+  "hostname-override": "{{ .MasterPrivateIP }}",
+  "cluster-cidr": "172.30.0.0/16",
+  "logtostderr": true,
+  "v": 0,
+  "allow-privileged": true,
+  "master": "http://{{ .MasterPrivateIP }}:8080",
```
saheienko: Why do you hardcode a master port?

scripts/start_kube_proxy.sh
```diff
@@ -0,0 +1,3 @@
...
+sudo docker run --privileged=true --volume=/etc/ssl/cer:/usr/share/ca-certificates --volume=/etc/kubernetes/worker-kubeconfig.yaml:/etc/kubernetes/worker-kubeconfig.yaml:ro --volume=/etc/kubernetes/ssl:/etc/kubernetes/ssl gcr.io/google_containers/hyperkube:v1.8.7 /hyperkube proxy --config /etc/kubernetes/config.json --master http://{{ .MasterPrivateIP }}
```
saheienko: Hardcoded hyperkube version...

pkg/jobs/digitalocean/node.go
```diff
@@ -0,0 +1,136 @@
+package digitalocean
```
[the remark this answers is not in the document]
stgleb: @saheienko they are intended to do so, but since we've got common yaml files for aws, do and os we can reuse it. Now it is done separately for digitalocean.

pkg/jobs/util.go
stgleb: @saheienko yeah

scripts/start_kube_proxy.sh
stgleb: @saheienko it i a shell scriptthat needs config file to start kube-proxy. Config is separate because it alsois template.

pkg/jobs/digitalocean/node.go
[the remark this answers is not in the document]
saheienko: @stgleb What the difference for installing a kubelet service on Ubuntu on AWS and DigitalOcean machine?

scripts/templates/config.json.tpl
stgleb: @saheienko I thought it is standard, previously we have a lot of hardcoded ports 8080 in yaml files.

saheienko: @stgleb That's not a standart, but default.

saheienko:
@yegorch Good job..

scripts/start_kubelet.sh.tpl
```diff
@@ -0,0 +1,40 @@
...
+              -v /var/run:/var/run:rw \
+              -v /var/lib/docker/:/var/lib/docker:rw \
+              -v /var/lib/kubelet/:/var/lib/kubelet:shared \
+              -v /var/log:/var/log:shared \
+              -v /srv/kubernetes:/srv/kubernetes:ro \
+              -v /etc/kubernetes:/etc/kubernetes:ro \
+              gcr.io/google-containers/hyperkube:v{{ .KubernetesVersion }} \
+              /hyperkube kubelet --allow-privileged=true \
+              --cluster-dns=10.3.0.10 \
+              --cluster_domain=cluster.local \
```
saheienko: @yegorch What about hardcoded vars?

pkg/jobs/node/node.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+type JobConfig struct {
+	MasterPrivateIP   string
+	ProxyPort         string
+	EtcdClientPort    string
+	KubernetesVersion string
+}
+
+func NewJob(startKubeletTemplate, startProxyTemplate *template.Template,
```
saheienko: @stgleb What component will be responsible for these templates?

stgleb: This should come from outside. Job has only one reason to change - the  job, but it should not be aware of how templates come in. Theoretically they can be read from network storage etc.

PR #520, opened 2018-06-08
Extend a kube service: add proxy getter for kubernetes resources. 

saheienko:
This change extends a functionality of a supergiant server.
It will be possible to:
- list all supported kubernetes resources;
- retrieve filtered list of the specified resources.


internal/kube/handler.go
```diff
@@ -0,0 +1,134 @@
...
+
+func (h *Handler) getKube(w http.ResponseWriter, r *http.Request) {
+	vars := mux.Vars(r)
+	if vars == nil {
+		http.Error(w, "invalid url path", http.StatusBadRequest)
+	}
+
+	id := vars["id"]
+	k, err := h.svc.Get(r.Context(), id)
+	if err != nil {
```
yegorch: What is kube is not found?

internal/kube/service.go
```diff
@@ -1,92 +1,171 @@
...
 	if err != nil {
-		return nil, errors.WithStack(err)
+		return nil, errors.Wrap(err, "marshal")
 	}
-	err = s.repository.Put(ctx, prefix, kube.Name, raw)
+
+	err = s.storage.Put(ctx, prefix, k.Name, raw)
 	if err != nil {
-		return nil, err
+		return nil, errors.Wrap(err, "storage")
```
yegorch: Nice error message - "storage"

internal/kube/service.go
```diff
@@ -1,92 +1,171 @@
...
 	}
 
-	logrus.Debug("kube.Service.ListAll end")
 	return kubes, nil
 }
 
-func (s *Service) Get(ctx context.Context, kubeName string) (*Kube, error) {
-	logrus.Debug("kube.Service.Get start")
-	rawJSON, err := s.repository.Get(ctx, prefix, kubeName)
+// Get deletes a kube with a specified name.
```
yegorch: Get Deletes a kube?

internal/kube/handler.go
saheienko: Yeap, i've missed it.

internal/kube/service.go
saheienko: It is an error prefix, the error will be "storage: err msg" and it will be finally wrapped with some context.
But.. BTW, it'd be better to use "storage: put" prefix here.

internal/kube/service.go
saheienko: Good catch :)

internal/kube/client.go
```diff
@@ -0,0 +1,78 @@
...
+	cfg, err := buildConfig(k.APIAddr, k.Auth)
+	if err != nil {
+		return nil, err
+	}
+
+	return discovery.NewDiscoveryClientForConfig(cfg)
+}
+
+// buildKubeConfig returns a kube config for provided options.
+func buildKubeConfig(addr string, auth Auth) clientcmddapi.Config {
```
yegorch: Any validations? Or it's handled by a library?

saheienko:
Squash it and reopen.

PR #523, opened 2018-06-12
Extend a kube service: add proxy getter for kubernetes resources.

saheienko:
This change extends a functionality of a supergiant server.
It will be possible to:
  -   list all supported kubernetes resources;
  -   retrieve filtered list of the specified resources.


pkg/kube/handler.go
```diff
@@ -5,76 +5,135 @@ import (
...
+}
+
+// NewHandler constructs a Handler for kubes.
+func NewHandler(svc *Service) *Handler {
+	return &Handler{svc}
+}
+
+// Register adds kube handlers to a router.
+func (h *Handler) Register(r *mux.Router) {
+	r.HandleFunc("/kubes", h.createKube).Methods(http.MethodPost)
```
stgleb: Why do you use plural name for actions over single kube?

pkg/kube/handler.go
```diff
@@ -5,76 +5,135 @@ import (
...
+	r.HandleFunc("/kubes", h.listKubes).Methods(http.MethodGet)
+	r.HandleFunc("/kubes/{kname}", h.getKube).Methods(http.MethodGet)
+	r.HandleFunc("/kubes/{kname}", h.deleteKube).Methods(http.MethodDelete)
```
stgleb: Why not just `/kubes/{kname}/resources/` ? list in url looks like not restful way.

pkg/kube/handler.go
```diff
@@ -5,76 +5,135 @@ import (
...
+
+	w.WriteHeader(http.StatusAccepted)
+}
+
+func (h *Handler) getKube(w http.ResponseWriter, r *http.Request) {
 	vars := mux.Vars(r)
-	kubeId := vars["id"]
-	kube, err := e.service.Get(r.Context(), kubeId)
+
+	kname := vars["kname"]
```
stgleb: Can you add ok variable for map access for failover?

pkg/kube/handler.go
```diff
@@ -5,76 +5,135 @@ import (
...
+	rs := vars["resource"]
+	ns := r.URL.Query().Get("namespace")
+	name := r.URL.Query().Get("resourceName")
 
-	if err := e.service.Delete(r.Context(), kubeId); err != nil {
-		logrus.Error(err)
-		http.Error(w, err.Error(), http.StatusInternalServerError)
+	rawResources, err := h.svc.GetKubeResources(r.Context(), kname, rs, ns, name)
+	if err != nil {
+		http.Error(w, "", http.StatusInternalServerError)
```
yegorch: SnedUnkownError here

pkg/kube/types.go
```diff
@@ -0,0 +1,40 @@
...
+
+// Kube represents a kubernetes cluster.
+type Kube struct {
+	Name              string `json:"name" valid:"required"`
+	Version           string
+	RBACEnabled       bool
+	AccountName       string
+	KubeProfileName   string
+	KubernetesVersion string
+	ProfileName       string
```
yegorch: Should be master profile name and node profile name as they might be different.

pkg/kube/handler.go
```diff
@@ -5,76 +5,135 @@ import (
...
 	}
 
-	if err = e.service.Create(r.Context(), kube); err != nil {
-		logrus.Error(err)
-		http.Error(w, err.Error(), http.StatusInternalServerError)
+	kname := vars["kname"]
+
+	rawResources, err := h.svc.ListKubeResources(r.Context(), kname)
+	if err != nil {
+		http.Error(w, "", http.StatusInternalServerError)
```
yegorch: message.SendUnknownError there

pkg/kube/handler.go
```diff
@@ -5,76 +5,135 @@ import (
...
-	ok, err := govalidator.ValidateStruct(kube)
-	if !ok {
-		http.Error(w, err.Error(), http.StatusBadRequest)
+	w.WriteHeader(http.StatusAccepted)
+}
+
+func (h *Handler) listResources(w http.ResponseWriter, r *http.Request) {
+	vars := mux.Vars(r)
+	if vars == nil {
+		http.Error(w, "invalid url path", http.StatusBadRequest)
```
yegorch: create and send message there using message.SendMessage func

pkg/kube/handler.go
```diff
@@ -5,76 +5,135 @@ import (
...
 }
 
-func (e *Handler) GetAllKubes(w http.ResponseWriter, r *http.Request) {
-	kubes, err := e.service.GetAll(r.Context())
+func (h *Handler) listKubes(w http.ResponseWriter, r *http.Request) {
+	kubes, err := h.svc.ListAll(r.Context())
 	if err != nil {
-		logrus.Error(err)
-		http.Error(w, err.Error(), http.StatusInternalServerError)
+		http.Error(w, "", http.StatusInternalServerError)
```
yegorch: message.SendUnknownError(w, err)

pkg/kube/handler.go
```diff
@@ -5,76 +5,135 @@ import (
...
+	vars := mux.Vars(r)
+	if vars == nil {
+		http.Error(w, "invalid url path", http.StatusBadRequest)
 		return
 	}
 
-	if err = e.service.Create(r.Context(), kube); err != nil {
-		logrus.Error(err)
-		http.Error(w, err.Error(), http.StatusInternalServerError)
+	kname := vars["kname"]
```
yegorch: What if kname is ""?

pkg/kube/handler.go
```diff
@@ -5,76 +5,135 @@ import (
...
+		message.SendUnknownError(w, err)
+	}
 }
 
-func (e *Handler) DeleteKube(w http.ResponseWriter, r *http.Request) {
+func (h *Handler) getResource(w http.ResponseWriter, r *http.Request) {
 	vars := mux.Vars(r)
-	kubeId := vars["id"]
+	if vars == nil {
+		http.Error(w, "invalid url path", http.StatusBadRequest)
```
yegorch: Send message there

pkg/kube/handler.go
```diff
@@ -5,76 +5,135 @@ import (
...
+		return
+	}
+
+	kname := vars["kname"]
+	rs := vars["resource"]
```
yegorch: Check if vars are ""

pkg/kube/handler.go
saheienko: Map access is safe here. (if the particular key won't found, the default value will be returned)

pkg/kube/handler.go
saheienko: Because it returns different outputs. Is it a restful way to serve a different resource on the same path?

pkg/kube/handler.go
saheienko: A kube won't be found.

pkg/kube/types.go
saheienko: Have no idea how it will be used. Removing it now.

pkg/kube/handler.go
saheienko: Updated.

pkg/kube/handler.go
saheienko: Updated.

pkg/kube/handler.go
saheienko: Updated.

pkg/kube/handler.go
saheienko: Updated.

PR #527, opened 2018-06-18
 Add certs handler for kube entity

saheienko:
This change adds an ability to retrieve a key/cert bundle from a kubernetes master.


pkg/kube/certs.go
```diff
@@ -0,0 +1,101 @@
...
+)
+
+// Bundler represents a key/certificate pair.
+type Bundle struct {
+	Cert []byte
+	Key  []byte
+}
+
+// Certs handles keys and certificates for kube components.
+type Certs struct {
```
yegorch: Is it possible to make it an interface to mock it out in tests?

pkg/kube/certs.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+
+	return &Bundle{
+		Cert: crtBytes,
+		Key:  crtKey,
+	}, nil
+}
+
+func (c *Certs) getFile(ctx context.Context, path string) ([]byte, error) {
+	stdout, stderr := &bytes.Buffer{}, &bytes.Buffer{}
```
yegorch: Using bash to read a file from go code is ... questionable practice. Please rewrite it using standard golang functions e.g. iotuil.ReadFile, etc.

pkg/kube/certs_test.go
```diff
@@ -0,0 +1,205 @@
...
+	"github.com/supergiant/supergiant/pkg/runner"
+	"github.com/supergiant/supergiant/pkg/runner/command"
+	"github.com/supergiant/supergiant/pkg/runner/shell"
+)
+
+var (
+	fakeErrFileNotFound = errors.New("key file not found")
+)
+
+type fakeRunner struct {
```
yegorch: Why not just outright mock it using testify mock package?

pkg/kube/certs.go
```diff
@@ -0,0 +1,101 @@
...
+		return nil, err
+	}
+
+	return &Bundle{
+		Cert: crtBytes,
+		Key:  crtKey,
+	}, nil
+}
+
+func (c *Certs) getFile(ctx context.Context, path string) ([]byte, error) {
```
yegorch: Can you please rename this method? Since it is intended to fetch certs from the k8s node the name of it is misleading as it's looks like it tries to read a local file using runner. 
Make it something like readCertificateFiles

stgleb: why you call these writers as std? this looks confusing since std is only for standard file descriptors that are passed to an application.

pkg/kube/certs.go
```diff
@@ -0,0 +1,101 @@
...
+	stdout, stderr := &bytes.Buffer{}, &bytes.Buffer{}
+	cmd := command.NewCommand(ctx, "cat", []string{path}, stdout, stderr)
+
+	err := c.r.Run(cmd)
+	if err != nil {
+		return nil, errors.Wrapf(err, "run %q", cmd.FullCommand())
+	}
```
stgleb: Are you sure that some output on stderr means that command was failed? Some application uses stderr as a default for log output(logrus).

yegorch: @stgleb :+1:  the only reliable way to determine if remote bash script has failed is to check it's exit code, if it's non zero - it's failed.

pkg/runner/runner.go
```diff
@@ -4,5 +4,5 @@ import "github.com/supergiant/supergiant/pkg/runner/command"
 
 // Runner is interface for running command in different environment
 type Runner interface {
-	Run(command command.Command)
```
stgleb: I would like to do that with separate pull request https://github.com/supergiant/supergiant/pull/528

pkg/kube/certs_test.go
saheienko: Its pretty hard to make it with a testify mock. You can update this tests later.

pkg/kube/service.go
```diff
@@ -151,6 +153,36 @@ func (s *Service) GetKubeResources(ctx context.Context, kname, resource, ns, nam
 	return raw, nil
 }
 
+// GetCerts returns a keys bundle for provided component name.
+func (s *Service) GetCerts(ctx context.Context, kname, cname string) (*Bundle, error) {
+	kube, err := s.Get(ctx, kname)
+	if err != nil {
+		return nil, errors.Wrap(err, "storage: get")
```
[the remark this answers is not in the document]
saheienko: It's already returns this error:
```
	raw, err := s.storage.Get(ctx, s.prefix, name)
	if err != nil {
		return nil, errors.Wrap(err, "storage: get")
	}
	if raw == nil {
		return nil, sgerrors.ErrNotFound
	}
```
Better to update a wrapper message...

pkg/kube/certs.go
saheienko: Nope, it's not intended for certificates only. This is a generic func and it can retrieve any file on the provided path.

pkg/kube/certs.go
saheienko: You are right. I've just actually used current implementations of the ssh client.
@yegorch If you have any ideas about retrieving an exit code here, you are welcome.

pkg/runner/runner.go
saheienko: Yeap, I'll sync with it.

pkg/kube/certs.go
stgleb: yes, but it can be named just out, because standard output is namely standard FD that is created per process.

PR #528, opened 2018-06-19
Refactor runner's code

stgleb:
Remove args from command
Change command name to s cript
Change Runner interface


stgleb:
@saheienko Sure, I will

stgleb:
@saheienko  I've looked once again. How do you want test looks like? Because SshRunner tightly coupled to ssh.Client there is no way to avoid direct calls to ssh. Do you want me to decouple client from SshRunner by interface?

saheienko:
```
+func (r *Runner) Run(c *command.Command) (err error) {
 	if r.client == nil {
 	if r.client == nil {
 		return errors.New("not connected")
 		return errors.New("not connected")
 	}
 	}
 
 
-	cmd := strings.TrimSpace(c.FullCommand())
+	cmd := strings.TrimSpace(c.Script)
 	if cmd == "" {
 	if cmd == "" {
 		return nil
 		return nil
 	}
 	}
```
Do you need a ssh.Client for tesging this block of code?

pkg/runner/command/command.go
```diff
@@ -3,31 +3,23 @@ package command
...
-	Command string
-	Args    []string
+	Script string
 
 	Out io.Writer
 	Err io.Writer
 }
 
-func NewCommand(ctx context.Context, cmd string, args []string, out, err io.Writer) *Command {
+func NewCommand(ctx context.Context, script string, out, err io.Writer) *Command {
```
saheienko: How do you plan to run this script?

stgleb: @yegorch could you elaborate on that?

stgleb: @saheienko We  just pass valid script where each command goes line by line and it is executed entirely.

PR #530, opened 2018-06-19
Initial UI Cleanup

josephkane:
1. tighten up authenticated http requests

2. remove dead '/login' routes and redirects

3. remove nodes count from dashboard

4. remove header from login page

5. format new cloud account forms w/ layout key and
hide openstack option for now

6. implement JSON paste for GCE service account key

### DIFF
--- a/cmd/ui/assets/src/app/shared/auth/authenticated-http-service.service.ts
+++ b/cmd/ui/assets/src/app/shared/auth/authenticated-http-service.service.ts
@@ -17,7 +17,7 @@ export class AuthenticatedHttpService extends Http {
     return super.request(url, options).catch((error: Response) => {
       if ((error.status === 401 || error.status === 403) && (window.location.href.match(/\?/g) || []).length < 2) {
         console.log('User session has expired... Redirect to login.');
-        this.router.navigate(['/login']);
+        this.router.navigate(['']);
         return Observable.throw(null);
       }
       return Observable.throw(error);
--- a/cmd/ui/assets/src/app/system/cloud-accounts/new-cloud-account/new-cloud-account.component.html
+++ b/cmd/ui/assets/src/app/system/cloud-accounts/new-cloud-account/new-cloud-account.component.html
@@ -1,4 +1,4 @@
-<div class="subcontext-form">
+<div class="new-cloud-account-wrapper">
   <header>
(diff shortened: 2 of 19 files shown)


cmd/ui/assets/src/app/shared/auth/authenticated-http-service.service.ts
```diff
@@ -17,7 +17,7 @@ export class AuthenticatedHttpService extends Http {
     return super.request(url, options).catch((error: Response) => {
       if ((error.status === 401 || error.status === 403) && (window.location.href.match(/\?/g) || []).length < 2) {
         console.log('User session has expired... Redirect to login.');
-        this.router.navigate(['/login']);
+        this.router.navigate(['']);
         return Observable.throw(null);
```
yegorch: Why throw null? Wouldn't it be better to create smth like NonFoundException and throw it instead?

cmd/ui/assets/src/app/system/cloud-accounts/new-cloud-account/new-cloud-account.component.html
```diff
@@ -1,4 +1,4 @@
-<div class="subcontext-form">
+<div class="new-cloud-account-wrapper">
```
TheKLARKEN: Great minds think alike! 😁

cmd/ui/assets/src/app/system/cloud-accounts/new-cloud-account/new-cloud-account.component.scss
```diff
@@ -1,3 +1,38 @@
...
+			border: 1px solid #ced4da;
+			border-radius: .125rem;
+			color: #495057;
+			height: 50vh;
+		}
+	}
+
+	button[type="submit"] {
+		font-size: 1.3em;
+		margin-top: 50px;
```
TheKLARKEN: Why the switching between pixels, em, and rem? It's been a while since I've done this, so I've forgotten where each is useful.

josephkane: @TheKLARKEN here I used ems so the font size is relative to the element's font size, so it will change if the user has it customized or cmd+/cmd-, and pixels for spacing so it is always in the same 'distance', regardless of zoom level. Im happen to change it

PR #531, opened 2018-06-21
aws provider: add CreateInstance and DeleteInstance methods.

saheienko:
Extract a CreateInstance and DeleteInstance method from the existing aws provider. Add new tests.
aws-skd package was updated to v1.14.6.


pkg/provider/v1/aws/interface.go
```diff
@@ -0,0 +1,10 @@
+package aws
+
+import (
+	"context"
+)
+
```
stgleb: What us the reason of creating this interface?

pkg/provider/v1/aws/instances.go
```diff
@@ -0,0 +1,152 @@
...
+	ErrNoInstancesCreated = errors.New("aws: no instances were created")
+)
+
+type provider struct {
+	session *session.Session
+
+	ec2SvcFn func(s *session.Session, region string) ec2iface.EC2API
+}
+
+// New returns a configured aws provider.
```
stgleb: Should we follow the principle "accept interface and return concrete type"?

saheienko: Do you have some objections with it? Suppose, case with ssh runner showed the advantages of this approach.
On the other hand, for what reason do you want to deal with struct instead of interface?

stgleb: Ok, runner interface was declared not specifically for ssh, but for runner at all with consideration that we can implement new runner. It declares methods that are needed to be runner. Also NewSshRunner returns concrete type instead of interface. Do you have a plan to extend this provider with other implementation? If yes - then interface should be declared in a root of provider package, but not for specific aws.

test/fake_aws_provider/fake_S3.go
```diff
@@ -3,9 +3,11 @@ package fake_aws_provider
 import (
```
yegorch: No need to update old mocks as we'll delete them anyway.

pkg/provider/v1/aws/instances.go
yegorch: Generally it is better to return concrete type in constructor. But in that particular case, since there would be only two implementations - real and mock, we can keep it.

saheienko: You are right. This interface initially was added as a generic for providers. But then, as InstanceConfig has grown, it was moved here instead of deletion.
Removing it.

pkg/provider/v1/aws/interface.go
saheienko: Originally it was created as a generic interface for providers. But now there is no need in it.

test/fake_aws_provider/fake_S3.go
saheienko: Yeap, but it has already done.

pkg/clouds/aws/instances.go
```diff
@@ -0,0 +1,152 @@
...
+// DeleteInstance terminates an instance with provided id and region.
+func (p *Provider) DeleteInstance(ctx context.Context, region, instanceID string) error {
+	if strings.TrimSpace(region) == "" {
+		return ErrNoRegionProvided
+	}
+	if strings.TrimSpace(instanceID) == "" {
+		return ErrInstanceIDEmpty
+	}
+	ec2S := p.ec2SvcFn(p.session, region)
+
```
stgleb: Why don't we pass trimmed strings there?

saheienko: Yeap, it could be better to replace vars with the trimmed ones.

PR #532, opened 2018-06-22
Refactor ssh runner package

saheienko:
* remove ssh.ClientConfig from Config parameters;
* refactor errors handling;
* update tests.

### DIFF
--- a/pkg/runner/ssh/runner.go
+++ b/pkg/runner/ssh/runner.go
@@ -5,97 +5,85 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/pkg/errors"
+
 	"golang.org/x/crypto/ssh"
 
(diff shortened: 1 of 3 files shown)


pkg/runner/ssh/runner.go
```diff
@@ -5,97 +5,85 @@ import (
...
-
-	err = session.Start(cmd)
-	if err != nil {
-		return err
-	}
+	session.Stdout = cmd.Out
+	session.Stderr = cmd.Err
 
-	waitCh := make(chan error)
+	waitCh := make(chan error, 1)
```
stgleb: I think we don't need this buffer.

saheienko: It is more reliable way to pass data through buffered channels.

PR #533, opened 2018-06-22
Add flannel task

stgleb:


pkg/runner/ssh/runner.go
```diff
@@ -74,7 +74,7 @@ func (r *Runner) Run(cmd *runner.Command) (err error) {
 	defer session.Close()
 
 	session.Stdout = cmd.Out
-	session.Stderr = cmd.Err
+	session.Stderr = cmd.Out
```
saheienko: Why did you change it?

scripts/flannel.sh.tpl
```diff
@@ -0,0 +1,15 @@
...
+mv /usr/bin/flanneld-{{ .Arch }} /usr/bin/flanneld
+chmod 755 /usr/bin/flanneld
+sed -i 's/REPLACEME/'`ifconfig|grep "10\."|grep "inet "|cut -f10 -d" "`'/g' /etc/default/flanneld
+
+echo "[Unit]
+Description=Networking service
+Requires=etcd-member.service
+[Service]
+Environment=FLANNEL_IMAGE_TAG=v{{ .Version }}
+ExecStartPre=/usr/bin/etcdctl set /coreos.com/network/config '{"Network":"{{ .Network }}", "Backend": {"Type": "{{ .NetworkType }}"}}'" > \
```
saheienko: Double quotes in the json config will be suppressed by the shell:
````
$ echo "[Unit]
Description=Networking service
Requires=etcd-member.service
[Service]
Environment=FLANNEL_IMAGE_TAG=v{{ .Version }}
ExecStartPre=/usr/bin/etcdctl set /coreos.com/network/config '{"Network":"{{ .Network }}", "Backend": {"Type": "{{ .NetworkType }}"}}'"

[Unit]
Description=Networking service
Requires=etcd-member.service
[Service]
Environment=FLANNEL_IMAGE_TAG=v{{ .Version }}
ExecStartPre=/usr/bin/etcdctl set /coreos.com/network/config '{Network:{{ .Network }}, Backend: {Type: {{ .NetworkType }}}}'
````

File: pkg/jobs/flannel.go [hunk 3]
```diff
@@ -0,0 +1,49 @@
...
+	"io"
+	"text/template"
+
+	"github.com/pkg/errors"
+
+	"github.com/supergiant/supergiant/pkg/runner"
+	"github.com/supergiant/supergiant/pkg/runner/ssh"
+)
+
+type FlannelJobConfig struct {
```
saheienko: It would be better to move this package under pkg/jobs/flannel to name this struct as Config.

scripts/flannel.sh.tpl
saheienko: Better to use cat in this case:
```
cat << EOF   
heredoc> [Unit]
Description=Networking service
Requires=etcd-member.service
[Service]
Environment=FLANNEL_IMAGE_TAG=v{{ .Version }}
ExecStartPre=/usr/bin/etcdctl set /coreos.com/network/config '{"Network":"{{ .Network }}", "Backend": {"Type": "{{ .NetworkType }}"}}'
EOF
```

pkg/jobs/flannel.go
stgleb: Ok

pkg/jobs/fake_runner.go
```diff
@@ -0,0 +1,23 @@
...
+	"errors"
+	"io"
+	"strings"
+
+
+	"github.com/supergiant/supergiant/pkg/runner"
+)
+
+type FakeRunner struct {
+	ErrMsg string
```
saheienko: Why do you you an error message instead of an error?

pkg/jobs/fake_runner.go
```diff
@@ -0,0 +1,23 @@
...
+import (
+	"errors"
+	"io"
+	"strings"
+
+
+	"github.com/supergiant/supergiant/pkg/runner"
+)
+
+type FakeRunner struct {
```
saheienko: Should it be part of the pkg/jobs package?

pkg/jobs/flannel/flannel_test.go
```diff
@@ -0,0 +1,107 @@
...
+/etc/systemd/system/flanneld.service
+systemctl enable flanneld.service
+systemctl restart flanneld.service
+`
+
+	tpl, err := template.New("").Parse(script)
+
+	if err != nil {
+		t.Errorf("error parsing flannel test template %s", err.Error())
+		return
```
saheienko: You can use [t.Fatalf()](https://golang.org/pkg/testing/#T.Fatalf) instead of t.Errorf() followed by return statement.

pkg/jobs/fake_runner.go
egopher: +1 Should be part of testutils

pkg/jobs/fake_runner.go
stgleb: I've found it easier to compare strings in tests instead of errors.

pkg/jobs/flannel/flannel_test.go
stgleb: @saheienko ok

pkg/jobs/fake_runner.go
saheienko: It is a bad practice, because different errors can have the same message.

stgleb: @saheienko Ok. I've changed that.

pkg/jobs/flannel/flannel.go
```diff
@@ -0,0 +1,50 @@
...
+	}, nil
+}
+
+func (i *Job) InstallFlannel(config JobConfig) error {
+	err := jobs.RunTemplate(context.Background(), i.scriptTemplate, i.runner, i.output, config)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
saheienko:
```
+	if err != nil {
+		return err
+	}
+
+	return nil
```
"gofmt -s" recognizes such blocks of code.

PR #542, opened 2018-06-27
WIP CMD for SG 2.0 + routes + fixes 

egopher:
* Created CMD for supergiant 2.0.
* Added API router.
* Added missing API register funcs
* Added missing constructor funcs for handlers and services
* Small Fixes


cmd/controlplane/main.go
```diff
@@ -0,0 +1,47 @@
...
+	"os"
+	"os/signal"
+	"syscall"
+
+	"github.com/sirupsen/logrus"
+	"github.com/supergiant/supergiant/pkg/controlplane"
+)
+
+var (
+	port         = flag.Int("port", 8080, "tcp port to listen for incoming requests")
```
stgleb: Using underscores in cmd args is a bit weird, we had better to use camel Case like most go projects or dash as most of unix utilities do.

cmd/controlplane/main.go
```diff
@@ -0,0 +1,47 @@
...
+	etcdURL      = flag.String("etcd_url", "localhost:2379", "etcd url with port")
+	templatesDir = flag.String("templates", "/etc/supergiant/templates", "supergiant will load script templates from the specified directory on start")
+	logLevel     = flag.String("log_level", "INFO", "logging level, e.g. info, warning, debug, error, fatal")
+)
+
+func main() {
```
stgleb: For the sake of convenience we can check env variables first. It will enhance user experience since you don't have to put cmd args each time you call the command.

cmd/controlplane/main.go
```diff
@@ -0,0 +1,47 @@
...
+
+	server, err := controlplane.New(cfg)
+	if err != nil {
+		logrus.Fatalf("broken configuration: %v", err)
+	}
+
+	sigs := make(chan os.Signal, 1)
+	signal.Notify(sigs, syscall.SIGINT, syscall.SIGTERM)
+
+	go func() {
```
stgleb: just

`<- sigs`

there is no need to placeholder here.

pkg/controlplane/server.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+func (srv *Server) Start() {
+	err := srv.server.ListenAndServe()
+	if err != nil {
+		logrus.Fatal(err)
+	}
+}
+
+func (srv *Server) Stop() {
```
stgleb: we should call shutdown to do that gracefull otherwise there is no big sense to make signal hanlder.

cmd/controlplane/main.go
saheienko: It is better to use hyphens (dashes) here.

pkg/controlplane/server.go
```diff
@@ -0,0 +1,148 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO add TLS support
+	s := &Server{
+		cfg: cfg,
+		server: http.Server{
+			Handler:      handlers.RecoveryHandler(handlers.PrintRecoveryStack(true))(r),
+			Addr:         fmt.Sprintf("0.0.0.0:%d", cfg.Port),
```
saheienko: It is better to use an explicit address.

cmd/controlplane/main.go
```diff
@@ -0,0 +1,47 @@
...
+)
+
+func main() {
+	flag.Parse()
+
+	cfg := &controlplane.Config{
+		Port:         *port,
+		EtcdUrl:      *etcdURL,
+		TemplatesDir: *templatesDir,
+		LogLevel:     *logLevel,
```
saheienko: Why do you pass LogLevel as a config param if logrus is used as global logger?

pkg/controlplane/server.go
```diff
@@ -0,0 +1,148 @@
...
+
+	return s, nil
+}
+
+func validate(cfg *Config) error {
+	if cfg.EtcdUrl == "" {
+		return errors.New("etcd url can't be empty")
+	}
+
+	if err := assert.CheckETCD(cfg.EtcdUrl); err != nil {
```
saheienko: Is it a validation step?

cmd/controlplane/main.go
egopher: We can, but let's do it in other PR.

cmd/controlplane/main.go
egopher: Fixed

PR #546, opened 2018-06-29
Fix unique constraint failure for integration tests

stgleb:


test/providers/testutils_test.go
```diff
@@ -113,7 +114,7 @@ func newServer() (*server.Server, error) {
 
 func createCloudAccount(client *client.Client, credentials map[string]string, provider string) (*model.CloudAccount, error) {
 	cloudAccount := &model.CloudAccount{
-		Name:        "test",
+		Name:        "test" + util.RandomString(4),
```
saheienko: It would be great to add provider info here.

PR #551, opened 2018-07-03
Fix a field name for helm releases

saheienko:
UI doesn't show a number of running apps, because uses an invalid field name for helm releases.


josephkane:
👍 LGTM

PR #553, opened 2018-07-03
Convert chart dates

josephkane:
* Display chart dates in an easier to read format

* dashboard, cluster-details, node and pod/resources details

### DIFF
--- a/cmd/ui/assets/src/app/shared/helpers/helpers.ts
+++ b/cmd/ui/assets/src/app/shared/helpers/helpers.ts
@@ -10,3 +10,14 @@ export function getBaseLocation() {
     return '/' + basePath;
   }
 }
+
+export function convertIsoToHumanReadable(isoString) {
+	let date = new Date(isoString);
+	let month = date.getMonth() + 1;
+	let day = date.getDay() + 1;
+	let fullYear = date.getFullYear();
+	let hours = date.getHours();
+	let minutes = date.getMinutes();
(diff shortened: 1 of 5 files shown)


aaronteague:
Looks pretty non-invasive.  Simply creating a function to take date in one format and spit it out to something a bit more elegant.  You have my approval.

cmd/ui/assets/src/app/shared/helpers/helpers.ts
```diff
@@ -10,3 +10,14 @@ export function getBaseLocation() {
...
+
+export function convertIsoToHumanReadable(isoString) {
+	let date = new Date(isoString);
+	let month = date.getMonth() + 1;
+	let day = date.getDay() + 1;
+	let fullYear = date.getFullYear();
+	let hours = date.getHours();
+	let minutes = date.getMinutes();
+	let seconds = date.getSeconds();
+	return `${month}-${day}-${fullYear}, ${hours}:${minutes}:${seconds}`;
```
TheKLARKEN: Just a thought, but do we need to tell people what year it is? This conversion is only in the UI, so the general "log-keeping" usefulness of seeing what year it is in every time entry is moot, and that slightly more compact stamp is a tad more elegant. I'd even argue that in 90% of cases where people need the graph, they don't even need the day or month, since you can't even zoom out to see "yesterday's spike in usage." Just wanted your thoughts. 😉

PR #565, opened 2018-07-17
Create tasks for task framework

stgleb:


scripts/systemd_update.sh.tpl
```diff
@@ -0,0 +1,45 @@
+    cat << EOF > /etc/systemd/system/{{ .KubeletsService }}
```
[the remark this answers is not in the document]
stgleb: Ok, I'll update names before moving from WIP.

scripts/kubernetes_post_start.sh.tpl
```diff
@@ -0,0 +1,19 @@
...
+/opt/bin/kubectl create clusterrolebinding system:dns-admin-binding --clusterrole=cluster-admin --user=system:dns
+/opt/bin/kubectl create clusterrolebinding add-ons-cluster-admin --clusterrole=cluster-admin --serviceaccount=kube-system:default
+/opt/bin/kubectl create clusterrolebinding default-user-cluster-admin --clusterrole=cluster-admin --user={{ .Username }}
+/opt/bin/kubectl create clusterrolebinding default-kube-system-admin --clusterrole=cluster-admin --serviceaccount=default:default --namespace=kube-system
+{{end}}
+
+/opt/bin/kubectl create -f /etc/kubernetes/addons/kube-dns.yaml
+/opt/bin/kubectl create -f /etc/kubernetes/addons/cluster-monitoring
+/opt/bin/kubectl create -f /etc/kubernetes/addons/default-storage-class.yaml
+/opt/bin/helm init
```
saheienko: You've already done in the install_tiller script

scripts/install_tiller.sh.tpl
```diff
@@ -0,0 +1,5 @@
+wget http://storage.googleapis.com/kubernetes-helm/{{ .HelmVersion }}-{{ .OperatingSystem }}-{{ .Arch }}.tar.gz --directory-prefix=/tmp/
```
saheienko: It's better to install tiller as an addon.

stgleb: @saheienko I'll probably need support for writing deployment for tiller, I'll do that in future, just don't want to get lots of different errors when assembling all this framework.

pkg/tasks/poststart/post_start.go
```diff
@@ -0,0 +1,53 @@
...
+		runner: sshRunner,
+		script: script,
+		output: outStream,
+	}
+
+	return t, nil
+}
+
+func (j *Task) PostStart(config Config) error {
+	err := tasks.RunTemplate(context.Background(), j.script, j.runner, j.output, config)
```
egopher: Better add context as an argument to the PostStart function, otherwise a user won't be able to pass timeout.

PR #569, opened 2018-07-24
Fixed password encryption issue

egopher:
Small fix as the password encryption wasn't invoked


pkg/user/service.go
```diff
@@ -27,6 +27,10 @@ func NewService(storagePrefix string, repository storage.Interface) *Service {
 }
 
 func (s *Service) Create(ctx context.Context, user *User) error {
+	err := user.encryptPassword()
```
saheienko: Please, check if user pointer is nil otherwise you'll get a panic.

PR #571, opened 2018-07-25
clouds/aws: add methods for retrieving ec2 types info

saheienko:


pkg/clouds/aws/instances.go
```diff
@@ -46,23 +52,48 @@ func New(keyID, secret string) (*Client, error) {
...
+		}
+
+		ec2Infos = append(ec2Infos, info)
+	}
+
+	return ec2Infos, nil
+}
+
+// CreateInstance starts a new instance due to the config.
+func (c *Client) CreateInstance(ctx context.Context, cfg InstanceConfig) (*ec2.Instance, error) {
```
egopher: Better make tags part of InstanceConfig

saheienko: Tags is a part of InstanceConfig.

build/gofmt.sh
```diff
@@ -13,9 +13,10 @@ find_files() {
 }
 
 GOFMT="gofmt -s"
-bad_files=$(find_files | xargs $GOFMT -l)
```
stgleb: Is it unrelated changes?

saheienko: Yeap, `make fmt` was broken by https://github.com/supergiant/supergiant/pull/587.

stgleb: Good, I also have noticed that. Maybe you can create separate PR for that so we can merge it ASAP. But it is not critical for me..

PR #574, opened 2018-07-26
Enable CORS

josephkane:
* re-enable CORS on the API
* expose auth response header for use with jwt

### DIFF
--- a/pkg/controlplane/server.go
+++ b/pkg/controlplane/server.go
@@ -60,12 +60,15 @@ func New(cfg *Config) (*Server, error) {
 	if err != nil {
 		return nil, err
 	}
+	headersOk := handlers.AllowedHeaders([]string{"Access-Control-Request-Headers", "Authorization"})
+	methodsOk := handlers.AllowedMethods([]string{"GET", "HEAD", "POST", "PUT", "OPTIONS"})
 
 	// TODO add TLS support
 	s := &Server{
 		cfg: cfg,
 		server: http.Server{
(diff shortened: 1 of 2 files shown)


pkg/user/handler.go
```diff
@@ -29,8 +29,17 @@ func NewHandler(userService *Service, tokenService *sgjwt.TokenService) *Handler
 	}
 }
```
stgleb: ResponseWriter is an interface it is reference type by itself you don't need to pass it by reference. https://research.swtch.com/interfaces

pkg/controlplane/server.go
```diff
@@ -60,12 +60,15 @@ func New(cfg *Config) (*Server, error) {
...
 		return nil, err
 	}
+	headersOk := handlers.AllowedHeaders([]string{"Access-Control-Request-Headers", "Authorization"})
+	methodsOk := handlers.AllowedMethods([]string{"GET", "HEAD", "POST", "PUT", "OPTIONS"})
 
 	// TODO add TLS support
 	s := &Server{
 		cfg: cfg,
 		server: http.Server{
-			Handler:      handlers.RecoveryHandler(handlers.PrintRecoveryStack(true))(r),
```
stgleb: You need to call function returned by handlers.CORS to wrap http.Handler from RecoveryHandler middleware

this will work

`handlers.CORS(headersOk, methodsOk)(handlers.RecoveryHandler(handlers.PrintRecoveryStack(true))(r))`

PR #575, opened 2018-07-27
Use coredns for a common/1.8 provider

saheienko:


saheienko:
Yeap, it should be for release 1.0 at first.

PR #579, opened 2018-07-31
Task api

stgleb:


pkg/workflows/workflow.go
```diff
@@ -24,99 +23,107 @@ import (
...
+type Workflow []steps.Step
+
+// Task is a workflow that runs, tracks it progress
+// workflow and task a related like binary file and process
+// in terms of operation system.
+type Task struct {
 	Id           string       `json:"id"`
 	Type         string       `json:"type"`
-	Config       steps.Config `json:"config"`
+	Config       steps.Config `json:"Cfg"`
```
saheienko: Why do you change it?

stgleb: accidentally while merging, I'll get this name back.

pkg/workflows/handler.go
```diff
@@ -30,3 +57,38 @@ func (h *WorkflowHandler) GetWorkflow(w http.ResponseWriter, r *http.Request) {
...
+	s := make([]steps.Step, 0, len(req.StepNames))
+	// Get steps for task
+	for _, stepName := range req.StepNames {
+		s = append(s, steps.GetStep(stepName))
+	}
+
+	task := New(s, req.Cfg, h.repository)
+	// We ignore cancel function since we cannot get it back
+	ctx, _ := context.WithTimeout(context.Background(), req.Cfg.Timeout*time.Second)
+	task.Run(ctx, os.Stdout)
```
saheienko: Why do you use os.Stdout here?

pkg/workflows/handler.go
```diff
@@ -5,23 +5,50 @@ import (
...
-	repository storage.Interface
+type TaskHandler struct {
+	runnerFactory func(config ssh.Config) (runner.Runner, error)
+	repository    storage.Interface
+}
+
+type BuildTaskRequest struct {
+	StepNames []string     `json:"step_names"`
+	Cfg       steps.Config `json:"config"`
+	SshConfig ssh.Config   `json:"ssh_config"`
```
saheienko: Please, use a camelCase format for json tags.

pkg/workflows/handler.go
stgleb: I use it like a mock, since we don't have support of websockets so far to stream output to client side.

pkg/workflows/steps/config.go
```diff
@@ -1,97 +1,99 @@
...
-	Name         string
-	K8sVersion   string
-	Region       string
-	Size         string
-	Role         string // master/node
-	Image        string
-	Fingerprints []string
-	AccessToken  string
+	Name         string   `json:"name"`
+	K8sVersion   string   `json:"k_8_s_version"`
```
saheienko: "k_8_s_version"?

stgleb: yeah, Goland autogeneration doesn't respect k8s

scripts/download_kubernetes_binary.sh.tpl
```diff
@@ -1,7 +1,6 @@
 #!/bin/bash
 source /etc/environment
-K8S_VERSION=v{{ .KubernetesVersion }}
 mkdir -p /opt/bin
-curl -sSL -o /opt/bin/kubectl https://storage.googleapis.com/kubernetes-release/release/${K8S_VERSION}/bin/{{ .OperatingSystem }}/{{ .Arch }}/kubectl
+curl -sSL -o /opt/bin/kubectl https://storage.googleapis.com/kubernetes-release/release/${K8S_VERSION}/bin/${OPERATING_SYSTEM}/${ARCH}/kubectl
```
TheKLARKEN: Why change these to environment variables but not the others? I was curious when to use environment variables and when to use Go variables. 🙂

stgleb: I've reconsidered that. I use go template variables as much as possible. Because I start new ssh session for each step and env variables will disappear.

pkg/workflows/workflow.go
```diff
@@ -24,99 +25,122 @@ import (
...
 	ErrMsg   string       `json:"errorMessage"`
 }
 
-type WorkFlow struct {
+// Workflow is a template for doing some actions
+type Workflow []steps.Step
+
+// Task is a workflow that runs, tracks it progress
+// workflow and task a related like binary file and process
+// in terms of operation system.
```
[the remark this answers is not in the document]
stgleb: Thanks

PR #585, opened 2018-08-14
Generate root user on cold start

egopher:


pkg/controlplane/server.go
```diff
@@ -11,6 +11,8 @@ import (
 	"github.com/gorilla/mux"
 	"github.com/sirupsen/logrus"
```
stgleb: goimpors

PR #587, opened 2018-08-14
Build process overhaul

egopher:
Instead of having multiple Dockerfiles there is one now which is using FROM SCRATCH and multi stage builds.


gopherstein:
Are we still cross-compiling for mac, windows, linux-x64, linux-arm64?

egopher:
@gopherstein we can do it, but we don't need that on every PR. To cross compile for different arch just do as follows:
`docker build -t supergiant . --build-arg ARCH=arm64`

I'll add cross compilation in makefiles, but for docker this is enough.

PR #597, opened 2018-08-27
Log streaming

stgleb:


pkg/provisioner/provisioner.go
```diff
@@ -28,6 +32,16 @@ func NewProvisioner(repository storage.Interface) *TaskProvisioner {
 		provisionMap: map[clouds.Name][]string{
 			clouds.DigitalOcean: {workflows.DigitalOceanMaster, workflows.DigitalOceanNode},
 		},
+		getWriter: func(name string) (io.WriteCloser, error) {
+			// TODO(stgleb): Add log directory to params of supergiant
+			f, err := os.OpenFile(path.Join("/tmp", name), os.O_CREATE|os.O_WRONLY|os.O_APPEND, 0666)
```
egopher: What if we have multiple simultaneous provisioning?

pkg/provisioner/provisioner.go
```diff
@@ -62,8 +76,20 @@ func (r *TaskProvisioner) Provision(ctx context.Context, kubeProfile *profile.Ku
...
 		// Provision master nodes
 		for _, masterTask := range masterTasks {
-			result := masterTask.Run(ctx, *config, os.Stdout)
-			err := <-result
+			if masterTask == nil {
+				continue
+			}
+
+			fdName := fmt.Sprintf("%s.log", masterTask.ID)
+			out, err := r.getWriter(fmt.Sprintf("%s.log", masterTask.ID))
```
saheienko: It's better to use a function for building `fdName`.

pkg/workflows/handler.go
```diff
@@ -163,3 +172,114 @@ func (h *TaskHandler) BuildAndRunTask(w http.ResponseWriter, r *http.Request) {
...
+		HandshakeTimeout: time.Second * 10,
+		WriteBufferSize:  1024,
+		ReadBufferSize:   0,
+	}
+
+	c, err := upgrader.Upgrade(w, r, nil)
+
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
```
saheienko: Please, log the error.

pkg/workflows/handler.go
```diff
@@ -163,3 +172,114 @@ func (h *TaskHandler) BuildAndRunTask(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	data, err := ioutil.ReadAll(fd)
+	c.SetWriteDeadline(time.Now().Add(time.Second * 10))
```
saheienko: Please, check the error.

pkg/workflows/handler.go
```diff
@@ -163,3 +172,114 @@ func (h *TaskHandler) BuildAndRunTask(w http.ResponseWriter, r *http.Request) {
...
+
+	err = c.WriteMessage(websocket.TextMessage, data)
+
+	if err != nil {
+		logrus.Errorf("Error while write to websocket %v", err)
+		return
+	}
+
+	go func(readCloser io.ReadCloser) {
+		pingTicker := time.NewTicker(time.Second * 60)
```
saheienko: Please, close this reader.

pkg/workflows/handler.go
```diff
@@ -163,3 +171,127 @@ func (h *TaskHandler) BuildAndRunTask(w http.ResponseWriter, r *http.Request) {
...
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	data, err := ioutil.ReadAll(fd)
+
+	if err != nil {
+		logrus.Errorf("Read content from file %s: %v", fd.Name(), err)
+		return
```
saheienko: What http response would be here?

pkg/provisioner/provisioner.go
```diff
@@ -28,6 +31,16 @@ func NewProvisioner(repository storage.Interface) *TaskProvisioner {
...
 		},
+		getWriter: func(name string) (io.WriteCloser, error) {
+			// TODO(stgleb): Add log directory to params of supergiant
+			f, err := os.OpenFile(path.Join("/tmp", name), os.O_CREATE|os.O_WRONLY|os.O_APPEND, 0666)
+
+			if err != nil {
+				return nil, err
+			}
+
+			return f, nil
```
saheienko: Redundant if statement here.

pkg/workflows/handler.go
stgleb: Nothing, when you do upgrade it makes a response with 101 code and Upgrade header.

pkg/workflows/handler.go
```diff
@@ -163,3 +171,108 @@ func (h *TaskHandler) BuildAndRunTask(w http.ResponseWriter, r *http.Request) {
...
+
+		for {
+			select {
+			case line := <-t.Lines:
+				c.SetWriteDeadline(time.Now().Add(time.Second * 10))
+				err = c.WriteMessage(websocket.TextMessage, []byte(line.Text))
+
+				// Do not log this error, since client can simply disconnect
+				if err != nil {
+					return
```
saheienko: It would be better to log the error.

pkg/workflows/handler.go
```diff
@@ -163,3 +171,108 @@ func (h *TaskHandler) BuildAndRunTask(w http.ResponseWriter, r *http.Request) {
...
+			Logger:      tail.DiscardingLogger,
+			MaxLineSize: 160,
+		})
+
+	if err != nil {
+		logrus.Errorf("Open file %s for tail %v", util.MakeFileName(id), err)
+		return
+	}
+
+	go func() {
```
saheienko: How do you close it when task is done?
Does it create the new ws connection for the same task?

pkg/workflows/handler.go
stgleb: This error is useless since it is just `Broken pipe` it gives no information about what happens on the wire since we operate with tcp connection. Error happens when client just closes browser tab so this  kind of error is completely ok, I just don't want to put these in logs. If we have many clients logs of SG will be polluted by this useless messages.

pkg/workflows/handler.go
stgleb: So, client has got a several options

`/tasks/<task_id>` for getting all task related info e.g step statuses

`/tasks/<task_id/logs` for getting task logs by websocket connection

`/task/<task_id>/restart` for restarting particular task if it fails.

Client side need polls task state and gets log updates by websocket. But once if task is failed it restarts it by api and refresh textarea for logs, there is no need to reestablish websocket connection. So This endpoints acts as log file watcher, rather  than task watcher.

pkg/workflows/handler.go
saheienko: So it returns the 101 code while failed to setup a ws handler (goroutine hasn't run)?

pkg/workflows/handler.go
```diff
@@ -3,22 +3,30 @@ package workflows
...
-
+	"github.com/gorilla/websocket"
+	"github.com/hpcloud/tail"
+	"github.com/sirupsen/logrus"
 	"github.com/supergiant/supergiant/pkg/runner"
 	"github.com/supergiant/supergiant/pkg/runner/ssh"
 	"github.com/supergiant/supergiant/pkg/storage"
+	"github.com/supergiant/supergiant/pkg/util"
 	"github.com/supergiant/supergiant/pkg/workflows/steps"
+	"io"
```
saheienko: fmt

pkg/workflows/handler.go
```diff
@@ -124,7 +138,15 @@ func (h *TaskHandler) RestartTask(w http.ResponseWriter, r *http.Request) {
...
-	task.Restart(context.Background(), id, os.Stdout)
+	writer, err := h.getWriter(id)
+
+	if err != nil {
+		http.Error(w, fmt.Sprintf("get writer %v", err), http.StatusInternalServerError)
+		logrus.Errorf("Get writer %v", err)
+		return
+	}
+
+	task.Restart(context.Background(), id, writer)
```
saheienko: Does it have no errors?

pkg/workflows/handler.go
```diff
@@ -163,3 +185,108 @@ func (h *TaskHandler) BuildAndRunTask(w http.ResponseWriter, r *http.Request) {
...
+				Offset: 0,
+				Whence: io.SeekStart,
+			},
+			Logger:      tail.DiscardingLogger,
+			MaxLineSize: 160,
+		})
+
+	if err != nil {
+		logrus.Errorf("Open file %s for tail %v", util.MakeFileName(id), err)
+		return
```
saheienko: It failed to setup a ws handler (goroutine hasn't been started), but returns the 101 http code?

pkg/workflows/handler.go
stgleb: This returns a channel with errors, we don't wait until it occures. We just return 202 Accepted and client can poll task state through the API.

pkg/workflows/handler.go
stgleb: gotcha

PR #616, opened 2018-09-06
Add save cluster method to provisioning

stgleb:


pkg/kube/client.go
```diff
@@ -8,10 +8,21 @@ import (
...
-	cfg, err := buildConfig(k.APIAddr, k.Auth)
+	var n *node.Node
+
+	if len(k.Masters) > 0 {
+		n = k.Masters[0]
+	} else {
+		return nil, sgerrors.ErrNotFound
+	}
+
+	cfg, err := buildConfig(n.PublicIp, k.Auth)
```
saheienko: You could simplify it:
```
	if len(k.Masters) == 0 {
		return nil, sgerrors.ErrNotFound
	}
	cfg, err := buildConfig(k.Masters[0].PublicIp, k.Masters[0].Auth)
```
And, it would be better to wrap `not found` error (what not found).

pkg/kube/client.go
```diff
@@ -21,7 +32,15 @@ func restClientForGroupVersion(k *Kube, gv schema.GroupVersion) (rest.Interface,
...
```
saheienko: Same as above.

pkg/kube/client.go
```diff
@@ -34,14 +53,14 @@ func buildKubeConfig(addr string, auth Auth) clientcmddapi.Config {
 	return clientcmddapi.Config{
 		AuthInfos: map[string]*clientcmddapi.AuthInfo{
 			auth.Username: {
-				Token:                 auth.Token,
+				Token: auth.Token,
```
saheienko: Please, run `make fmt`.

pkg/kube/types.go
```diff
@@ -20,3 +33,10 @@ type Auth struct {
 	Cert     string `json:"cert"`
 	Key      string `json:"key"`
 }
+
+type Networking struct {
+	Manager string `json:"networkManager"`
+	Version string `json:"flannelVersion"`
+	Type    string `json:"networkType"`
```
saheienko: Please, update json tags too.

pkg/provisioner/provisioner.go
```diff
@@ -101,6 +104,9 @@ func (r *TaskProvisioner) Provision(ctx context.Context, profile *profile.Profil
 		// Wait for cluster checks are finished
 		r.waitCluster(ctx, clusterTask, config)
 
+		logrus.Info("Save cluster")
```
saheienko: Please, add some info about a cluster to the log message.

pkg/kube/client.go
stgleb: I ran it a few times, for some reason this is not changed.

PR #650, opened 2018-09-14
Add node to cluster

stgleb:
Implement add node to cluster that was previously
deployed


pkg/kube/handler.go
```diff
@@ -208,3 +225,135 @@ func (h *Handler) getCerts(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	writer, err := getWriter(t.ID)
+
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	errChan := t.Run(context.Background(), *config, writer)
```
egopher: Better use context from the request.

stgleb: Not in that case, since request will be finished much earlier that task.

pkg/kube/handler.go
```diff
@@ -208,3 +225,135 @@ func (h *Handler) getCerts(w http.ResponseWriter, r *http.Request) {
...
+	// Respond to client side that request has been accepted
+	w.WriteHeader(http.StatusAccepted)
+	err = json.NewEncoder(w).Encode(&workflows.TaskResponse{
+		ID: t.ID,
```
egopher: Is it possible to add an URL that could be polled?

stgleb: Sure, as a good rest api is will be nice to a dd links field

pkg/kube/handler.go
```diff
@@ -208,3 +229,111 @@ func (h *Handler) getCerts(w http.ResponseWriter, r *http.Request) {
...
+		http.NotFound(w, r)
+		return
+	}
+
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	kubeProfile := profile.Profile{
```
egopher: Do we really need all this info? I mean cluster is already created.

stgleb: Yes, I need all these settings because simple node creation needs all of these.

PR #653, opened 2018-09-17
Revert ng 6 update

nikage:


pkg/workflows/utils.go
```diff
@@ -77,3 +78,11 @@ func DeserializeTask(data []byte, repository storage.Interface) (*Task, error) {
...
 	return task, nil
 }
+
+func MakeKeyName(name string, isUser bool) string {
+	if isUser {
+		return fmt.Sprintf("%s-user", name)
+	}
+
+	return fmt.Sprintf("%s-provision", name)
+}
```
josephkane: what happened here? just curious as to why this is part of a frontend revert?

nikage: I've accidentally squashed @gleb's commit but I reset this. It's better the @stgleb or @egopher check this in addition

PR #654, opened 2018-09-17
Delete node from cluster

stgleb:


pkg/provisioner/provisioner_test.go
```diff
@@ -30,13 +39,18 @@ func TestProvisionCluster(t *testing.T) {
 	repository := &testutils.MockStorage{}
 	repository.On("Put", context.Background(), mock.Anything, mock.Anything, mock.Anything).Return(nil)
 
+	bc := &bufferCloser{
```
egopher: Why not defer?

pkg/util/util.go
```diff
@@ -321,3 +324,15 @@ func FillCloudAccountCredentials(ctx context.Context, getter cloudAccountGetter,
 
 	return nil
 }
+
+func GetAnyNode(nodeMap map[string]*node.Node) *node.Node {
```
egopher: More like get random node

stgleb: Random assumes that it will be random each time, since for map it is predictable when map is not changed,

pkg/provisioner/provisioner_test.go
stgleb: What do you mean?

pkg/util/util.go
saheienko:
```
n = util.GetAnyNode(k.Masters)
```
Does it used as a random node getter?

pkg/kube/handler.go
```diff
@@ -349,3 +360,105 @@ func (h *Handler) addNode(w http.ResponseWriter, r *http.Request) {
...
+		if sgerrors.IsNotFound(err) {
+			message.SendNotFound(w, kname, err)
+			return
+		}
+		message.SendUnknownError(w, err)
+		return
+	}
+
+	if _, ok := k.Masters[nodeName]; ok {
+		http.Error(w, "delete master node not allowed", http.StatusMethodNotAllowed)
```
saheienko: Probably it should depend on a number of master nodes.

pkg/util/util.go
stgleb: no, it doesn't. It just picks first node from map.

egopher: Range in golang map is randomized.

pkg/workflows/steps/digitalocean/common.go
```diff
@@ -0,0 +1,42 @@
...
+type TagService interface {
+	TagResources(string, *godo.TagResourcesRequest) (*godo.Response, error)
+}
+
+type KeyService interface {
+	Create(context.Context, *godo.KeyCreateRequest) (*godo.Key, *godo.Response, error)
+}
+
+type DeleteService interface {
+	DeleteByTag(context.Context, string) (*godo.Response, error)
```
saheienko: Should it be a part of DropletService?

pkg/workflows/steps/digitalocean/common.go
```diff
@@ -0,0 +1,42 @@
...
+
+	"context"
+	"errors"
+	"github.com/digitalocean/godo"
+	"github.com/supergiant/supergiant/pkg/workflows/steps"
+)
+
+const (
+	CreateInstanceStepName = "createInstanceDigitalOcean"
+	DeleteInstanceStepName = "deleteInstanceDigitalOcean"
```
saheienko: It would be better to use `machine` instead of `instance`.

pkg/kube/handler.go
stgleb: Basically yes, but meanwhile we decided to move master node deletetion out of scope of mvp

pkg/workflows/steps/digitalocean/common.go
stgleb: I use subset of DropletService methods so this is easier to mock it.

pkg/workflows/steps/digitalocean/common.go
stgleb: Ok

pkg/kube/handler_test.go
```diff
@@ -35,6 +37,19 @@ type mockNodeProvisioner struct {
 	mock.Mock
 }
 
+type mockRepository struct {
```
egopher: We already that struct in testutils package.

stgleb: Good point, thanks

PR #655, opened 2018-09-20
Add cluster state

stgleb:
Track cluster and node state changes in separate
goroutine.


pkg/node/node.go
```diff
@@ -6,21 +6,36 @@ import (
...
+
+const (
+	StatePlanned      NodeState = "planned"
+	StateBuilding     NodeState = "building"
+	StateProvisioning NodeState = "provisioning"
+	StateError        NodeState = "error"
+	StateActive       NodeState = "active"
+
+	RoleMaster Role = "master"
+	RoleNode   Role = "node"
```
saheienko: Probably, it should be `worker`.

pkg/node/node.go
```diff
@@ -6,21 +6,36 @@ import (
...
 
+type NodeState string
+
+type Role string
+
+const (
+	StatePlanned      NodeState = "planned"
+	StateBuilding     NodeState = "building"
+	StateProvisioning NodeState = "provisioning"
+	StateError        NodeState = "error"
```
saheienko: Does an `error` state is self descriptive?

pkg/provisioner/provisioner_test.go
```diff
@@ -16,6 +16,7 @@ import (
 	"github.com/supergiant/supergiant/pkg/testutils"
 	"github.com/supergiant/supergiant/pkg/workflows"
 	"github.com/supergiant/supergiant/pkg/workflows/steps"
+	"time"
```
saheienko: Please, run `make fmt`.

pkg/node/node.go
stgleb: Not that much, more descriptive error is stored in corresponding task.

pkg/provisioner/provisioner_test.go
stgleb: my fault, even though`make fmt` doesn't heal this

pkg/node/node.go
stgleb: I think we can discuss and set up naming convention for entities while we are all here, I am open to any propositions.

PR #656, opened 2018-09-20
Delete cluster for DigitalOcean

stgleb:


pkg/kube/handler.go
```diff
@@ -195,6 +199,68 @@ func (h *Handler) deleteKube(w http.ResponseWriter, r *http.Request) {
...
+		}
+
+		message.SendUnknownError(w, err)
+		return
+	}
+
+	t, err := workflows.NewTask(h.workflowMap[acc.Provider].DeleteCluster, h.repo)
+
+	if err != nil {
+		if sgerrors.IsNotFound(err) {
```
egopher: Why is it not found? Maybe it's more like internal error?

pkg/kube/handler.go
```diff
@@ -195,6 +199,68 @@ func (h *Handler) deleteKube(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	writer, err := h.getWriter(t.ID)
+
+	if err != nil {
+		message.SendUnknownError(w, err)
+		return
+	}
+
+	errChan := t.Run(context.Background(), *config, writer)
```
egopher: Move that to the service.

pkg/kube/handler.go
stgleb: NewTask looks for appropriate workflow and it can return not found https://github.com/supergiant/supergiant/blob/master/pkg/workflows/task.go#L37

pkg/workflows/workflow.go
```diff
@@ -36,11 +36,19 @@ const (
 
 	Cluster = "Cluster"
 
-	DigitalOceanMaster     = "DigitalOceanMaster"
-	DigitalOceanNode       = "DigitalOceanNode"
-	DigitalOceanDeleteNode = "DigitalOceanDeleteNode"
+	DigitalOceanMaster        = "DigitalOceanMaster"
+	DigitalOceanNode          = "DigitalOceanNode"
```
saheienko: What the difference between `master` and `node` on DigitalOcean?

pkg/kube/handler.go
stgleb: I've considered service as layer that just modifies single entity like Kube, Node, Account etc.
But tassk framework and tassk itself is like a separate service and I don't think this is a good idea to call one service from another.

pkg/workflows/workflow.go
stgleb: Flows are pretty different https://github.com/supergiant/supergiant/blob/master/pkg/workflows/workflow.go#L52

PR #670, opened 2018-10-02
Fix overlay networking in SG cluster

stgleb:


templates/kubelet.tpl
```diff
@@ -27,7 +27,8 @@ ExecStart=/usr/bin/docker run \
       --cluster_domain=cluster.local \
       --pod-manifest-path=/etc/kubernetes/manifests \
       --kubeconfig=/etc/kubernetes/worker-kubeconfig.yaml \
-      --volume-plugin-dir=/etc/kubernetes/volumeplugins --fail-swap-on=false --register-node=true
+      --volume-plugin-dir=/etc/kubernetes/volumeplugins --fail-swap-on=false --register-node=true \
+      -v=6
```
saheienko: That's will print a pretty detailed log output. Do we need this by default?

stgleb: basically not, thank you for pointing that.

PR #671, opened 2018-10-03
Securing cluster with TLS

stgleb:
Fix creation of CA cert in supergiant
Distribute CA cert across nodes
Configure API server and kubelet to use TLS


templates/poststart.tpl
```diff
@@ -16,6 +16,12 @@ echo "PostStart started"
     {{end}}
 {{ else }}
     until $([ $(docker ps |grep hyperkube| wc -l) -eq 2 ]); do printf '.'; sleep 5; done
+
+    kubectl config set-cluster default-cluster --server="{{ .Host }}:{{ .Port }}"
+    kubectl config set-context default-system --cluster=default-cluster --user=default-admin
+    kubectl config use-context default-system
+
+    kubectl --kubeconfig /etc/kubernetes/worker-kubeconfig.yaml config set-credentials kubelet --client-certificate /etc/kubernetes/ssl/worker.pem --client-key /etc/kubernetes/ssl/worker-key.pem --server=https://{{ .Host }}
```
saheienko: `set-credentials` subcommand has no `--server` flag, use `set-cluster` for this.

pkg/workflows/steps/config.go
```diff
@@ -80,10 +85,10 @@ type NetworkConfig struct {
 }
 
 type KubeletConfig struct {
-	MasterPrivateIP string `json:"masterPrivateIP"`
-	ProxyPort       string `json:"proxyPort"`
-	EtcdClientPort  string `json:"etcdClientPort"`
-	K8SVersion      string `json:"k8sVersion"`
+	IsMaster       bool   `json:"isMaster"`
+	ProxyPort      string `json:"proxyPort"`
+	EtcdClientPort string `json:"etcdClientPort"`
```
saheienko: `kubelet` doesn't use etcd, why does this has `ETCDClientPort` config parameter?

pkg/workflows/steps/certificates/certificates_test.go
```diff
@@ -54,12 +55,20 @@ func TestWriteCertificates(t *testing.T) {
 
 	output := new(bytes.Buffer)
 
+	PKIBundle, err := pki.NewPKI(nil)
```
saheienko: `pki` couldn't be a bundle, probably it's better to rename `NewPKI` to `NewCA`.

templates/poststart.tpl
stgleb: flags: Specifies optional flags. For example, you can use the -s or --server flags to specify the address and port of the Kubernetes API server. Maybe it need to go before set-crede credentail command itself?

pkg/workflows/steps/config.go
stgleb: Remove it

gopherstein:
Conflict.

PR #685, opened 2018-10-16
Bugfix for step/task status reporting. 

egopher:


pkg/workflows/task.go
```diff
@@ -190,7 +199,7 @@ func (w *Task) startFrom(ctx context.Context, id string, out io.Writer, i int) e
 }
 
 // synchronize state of workflow to storage
-func (w *Task) sync(ctx context.Context) error {
```
stgleb: Do you plan to use this somewhere outside of task framework?

PR #688, opened 2018-10-17
Add flannel IP to certificate altnames

stgleb:


templates/certificates.tpl
```diff
@@ -16,7 +18,7 @@ DNS.1 = kubernetes
 DNS.2 = kubernetes.default
 DNS.3 = kubernetes.default.svc
 DNS.4 = kubernetes.default.svc.cluster
-IP.1 = {}
+IP.1 = {FLANNELIP}
```
saheienko: What is `FLANNELIP` used for?

stgleb: Placeholder for sed

PR #690, opened 2018-10-17
App store@enhancements

nikage:


cmd/ui/assets/package.json
```diff
@@ -12,7 +12,6 @@
   },
   "private": true,
   "dependencies": {
-    "@angular/animations": "^6.1.7",
```
josephkane: may need to rebase, the new provisioning visualization uses this

PR #699, opened 2018-10-19
Add kubernetes addr to altnames

stgleb:


templates/certificates.tpl
```diff
@@ -21,6 +21,7 @@ DNS.3 = kubernetes.default.svc
 DNS.4 = kubernetes.default.svc.cluster
 IP.1 = {MASTER_HOST}
 IP.2 = {PRIVATE_HOST}
+IP.3 = 10.3.0.1
```
saheienko: This is not a dns address:
```
templates/kubelet.tpl:27:      --cluster-dns=10.3.0.10 \
```
```
kubectl get svc
NAMESPACE     NAME            TYPE        CLUSTER-IP   EXTERNAL-IP   PORT(S)     AGE
default       kubernetes      ClusterIP   10.3.0.1     <none>        443/TCP     11h
```

PR #700, opened 2018-10-19
Increase test coverage

stgleb:


pkg/account/handler.go
```diff
@@ -15,6 +15,8 @@ import (
 	"github.com/supergiant/supergiant/pkg/sgerrors"
 )
 
+var SupportedMethods = [3]string{http.MethodGet, http.MethodPost, http.MethodDelete}
```
saheienko: Where does it used?

pkg/kube/handler_test.go
```diff
@@ -24,6 +24,7 @@ import (
 	"github.com/supergiant/supergiant/pkg/testutils"
 	"github.com/supergiant/supergiant/pkg/workflows"
 	"github.com/supergiant/supergiant/pkg/workflows/steps"
+	"strings"
```
saheienko: Please, run `make fmt`.

PR #713, opened 2018-10-25
Check cluster for duplicate

stgleb:


pkg/kube/handler.go
```diff
@@ -143,6 +143,11 @@ func (h *Handler) createKube(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	if kube2, _ := h.svc.Get(r.Context(), k.Name); kube2 != nil {
```
taaraora: My be it will be more meaningful to rename `k` as `newKube` and `kube2` as `existingKube` ?

PR #715, opened 2018-10-26
Add prometheus step

stgleb:


templates/prometheus.sh.tpl
```diff
@@ -0,0 +1,4 @@
+wget https://raw.githubusercontent.com/kubernetes/kubernetes/release-1.12/plugin/pkg/auth/authorizer/rbac/bootstrappolicy/testdata/cluster-roles.yaml
+sudo kubectl create -f cluster-roles.yaml --validate=false
```
saheienko: What if rbac is disabled?
```
pkg/model/kube.go:20:	RBACEnabled  bool      `json:"rbacEnabled"`
```

pkg/workflows/steps/prometheus/prometheus.go
```diff
@@ -0,0 +1,56 @@
...
+)
+
+const StepName = "prometheus"
+
+type Step struct {
+	script *template.Template
+}
+
+func (s *Step) Rollback(context.Context, io.Writer, *steps.Config) error {
+	return nil
```
taaraora: rollback was not supposed to be implemented?

templates/prometheus.sh.tpl
taaraora: ```sudo kubectl create -f cluster-roles.yaml``` looks like it is candidate for separate task?
Or may be it is better to use ```kubectl apply -f``` instead?

stgleb: Well it won't work, helm chart requires cluster-admin role

stgleb: Good point!

pkg/workflows/steps/prometheus/prometheus.go
stgleb: Rollback is not applicable for some sorts of steps. I don't think that it makes sense for this one.

stgleb: Rollback is not applicable for some sorts of steps. I don't think that it makes sense for this one.

templates/prometheus.sh.tpl
stgleb: Btw, what if we enable RBAC by default without option to turn it off? Will it break some kinds of workload?

saheienko: There is no authorization by default:
```
--authorization-mode stringSlice     Default: [AlwaysAllow]
--
  | Ordered  list of plug-ins to do authorization on secure port. Comma-delimited  list of: AlwaysAllow,AlwaysDeny,ABAC,Webhook,RBAC,Node.
```

to enable RBAC this should be added (apiserver has no config for it):
```
--authorization-mode=RBAC
```

PR #719, opened 2018-10-29
Add an icon link for the chart list items

saheienko:
While listing charts there was no info for the chart icon, add it to the ChartInfo struct.

### DIFF
--- a/pkg/sghelm/service.go
+++ b/pkg/sghelm/service.go
@@ -231,7 +231,7 @@ func toRepoInfo(e *repo.Entry, index *repo.IndexFile) *model.RepositoryInfo {
 			continue
 		}
 
-		// ensure charts are sorted in descending order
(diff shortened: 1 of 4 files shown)


pkg/sghelm/service.go
```diff
@@ -243,18 +243,26 @@ func toRepo(e *repo.Entry, index *repo.IndexFile) *model.RepositoryInfo {
...
 			Repo:        e.Name,
+			Icon:        iconFrom(entry),
 			Description: descriptionFrom(entry),
 			Versions:    toChartVersions(entry),
 		})
 	}
 	return r
 }
 
+func iconFrom(cvs repo.ChartVersions) string {
```
stgleb: Could you cover this with unit test?

saheienko: Done.

PR #721, opened 2018-10-30
No Cloud Accounts Prompt

josephkane:
Prompt users to create cloud accounts when:

* viewing cloud accounts list

* spinning up cluster


nikage:
LGTM

PR #739, opened 2018-11-05
Readme@update start command

nikage:


josephkane:
LGTM

nikage:
@gopherstein there are to many commits because I had checked out new branch not from my forked repo but from the original repo and those a are commits that hadn't been in the forked repo. 

If take a look on the changed files there is a single file README has changed.
If that's OK then unblock the PR please.

PR #744, opened 2018-11-06
App store@layout bugs

nikage:


cmd/ui/assets/package.json
```diff
@@ -12,7 +12,6 @@
   },
   "private": true,
   "dependencies": {
-    "@angular/animations": "^6.1.7",
```
josephkane: this will break the the task table on the provisioning visualizer. is there a way to stop this from happening?

PR #752, opened 2018-11-07
Check kube for duplicate

stgleb:


pkg/provisioner/handler.go
```diff
@@ -69,6 +75,11 @@ func (h *Handler) Provision(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	if existingKube, _ := h.kubeGetter.Get(r.Context(), req.ClusterName); existingKube != nil {
```
saheienko: Check an error, you can override the existing one in case of connection issues.

stgleb: Done

PR #753, opened 2018-11-07
Check cloud account for duplicate

stgleb:


pkg/account/handler.go
```diff
@@ -51,6 +51,18 @@ func (h *Handler) Create(rw http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	existingAccount, err := h.service.Get(r.Context(), account.Name)
+
+	if existingAccount != nil {
+		message.SendAlreadyExists(rw, account.Name, sgerrors.ErrAlreadyExists)
```
saheienko: Could you move the existence check to the service?

saheienko: Or move it to the `Create` method. Otherwise add a comment, that it won't check for existing ones.

PR #754, opened 2018-11-07
Added check to not add default repositories if not a first launch

egopher:


pkg/controlplane/server.go
```diff
@@ -278,6 +278,17 @@ func ensureHelmRepositories(svc sghelm.Servicer) {
 		return
 	}
 
+	repos, err := svc.ListRepos(context.Background())
+	if err != nil {
+		logrus.Errorf("failed to read repositories from database, please check if the database is running")
+		return
+	}
+
+	if len(repos) != 0 {
```
saheienko: What if all the repositories would be deleted?

PR #756, opened 2018-11-07
Add deleting state to node

stgleb:


pkg/kube/handler.go
```diff
@@ -521,7 +521,17 @@ func (h *Handler) deleteNode(w http.ResponseWriter, r *http.Request) {
 
 	// Update cluster state when deletion completes
 	go func() {
-		err := <-errChan
+		// Set node to deleting state
+		nodeToDelete := k.Nodes[nodeName]
+		nodeToDelete.State = node.StateDeleting
```
saheienko: Here is a possible panic, check if node exists in the map.

PR #761, opened 2018-11-08
fix (apps-list.component.html)  set black pixel instead of empty src

nikage:
- empty src caused stream of errors and cpu load


josephkane:
LGTM

PR #764, opened 2018-11-08
Change metric per node format

stgleb:
API endpoint

GET `/v1/api/kubes/{kube_name}/nodes/metrics`

Response body

```
{"master-node":{"cpu":"0.06149999999999989","memory":"0.431503826067256"},"worker-node":{"cpu":"0.0431666666666668","memory":"0.2798653943056244"}}
```

### DIFF
--- a/pkg/kube/handler.go
+++ b/pkg/kube/handler.go
@@ -8,7 +8,6 @@ import (
 	"io"
 	"net/http"
 	"strconv"
-	"strings"
 	"time"
 
 	"github.com/gorilla/mux"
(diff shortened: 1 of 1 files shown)


egopher:
Maybe it's better to floor it it two-three digits?

stgleb:
@egopher It is floating point number, when you display it will always have tail from 99999999999 or something else. It won't make any difference when display on the UI.

josephkane:
its trivial for the ui to handle this, not a big deal

egopher:
@josephkane @stgleb Better to handle this in UI trough...

PR #792, opened 2018-11-09
Remove Unsupported Elements

josephkane:
* remove preferences and logs from system menu

* remove unsupported cloud providers


eldridgeh:
LGTM

PR #801, opened 2018-11-09
Remove Provider Refs

josephkane:
* remove all references to unsupported providers


josephkane:
fix for https://github.com/supergiant/supergiant/issues/799 and https://github.com/supergiant/supergiant/issues/798

Issue #788, opened 2018-11-08
2.0: AWS - Unable to ssh into AWS boxes


eldridgeh:
Summary of Issue:
After creating a cluster on AWS, I am unable to ssh into the box.

Steps to Reproduce:
1. Spin up a cluster in AWS
2. Try to ssh into the master or a node.

Expected Results:
You are able to ssh into a box.

Actual Results:
You are unable to ssh in due to not having the PEM file.

Dev Info: (fill out and add links to log files)
1. `go version`: 1.10
2. SG latest commit hash or release tag: dddedff38267d34e16763fde1955ca4dec1cc5d7
3. Number of Masters and Nodes: Any
4. cloud provider: AWS

gopherstein:
Fixed with https://github.com/supergiant/supergiant/pull/804

PR #804, opened 2018-11-09
AWS Fixes

### LINKED ISSUE #780: 2.0: AWS - Not all machine types are shown
Summary of Issue:
When creating a cluster in AWS, you are unable to choose m4's etc.

Steps to Reproduce:
1. Start creating a cluster using us-west-1.
2. When on the machine type, find the m4

Expected Results:
All available machine types in the region are available.

Actual Results:
You are missing many instance types.

Dev Info: (fill out and add links to log files)
1. `go version`: 1.10
2. SG latest commit hash or release tag: dddedff38267d34e16763fde1955ca4dec1cc5d7
3. Number of Masters and Nodes: Any
4. cloud provider: AWS

egopher:
Fixes #780 #788 #770

### DIFF
--- a/pkg/workflows/steps/config.go
+++ b/pkg/workflows/steps/config.go
@@ -61,7 +61,6 @@ type AWSConfig struct {
 	ImageID                string `json:"image"`
 	InstanceType           string `json:"size"`
 	HasPublicAddr          string `json:"hasPublicAddr"`
-	LogPrivateKey          bool   `json:"logPrivateKey"`
 }
 
 type FlannelConfig struct {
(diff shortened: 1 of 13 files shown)


pkg/workflows/steps/config.go
```diff
@@ -294,9 +295,10 @@ func NewConfig(clusterName, discoveryUrl, cloudAccountName string, profile profi
 			Arch:            profile.Arch,
 		},
 		SshConfig: SshConfig{
-			Port:    "22",
-			User:    "root",
-			Timeout: 10,
+			Port:      "22",
+			User:      "root",
+			Timeout:   10,
```
stgleb: Why not to take it from cloudaccount?

pkg/workflows/workflow.go
```diff
@@ -100,6 +100,7 @@ func Init() {
 	awsMasterWorkflow := []steps.Step{
 		steps.GetStep(amazon.StepNameCreateEC2Instance),
 		steps.GetStep(ssh.StepName),
```
stgleb: Plese, move this step to separate package, since certificates and keys are  not related

pkg/workflows/workflow.go
```diff
@@ -115,13 +116,15 @@ func Init() {
...
+		steps.GetStep(certificates.AddAuthorizedKeyStepName),
 		steps.GetStep(downloadk8sbinary.StepName),
 		steps.GetStep(docker.StepName),
 		steps.GetStep(certificates.StepName),
 		steps.GetStep(manifest.StepName),
 		steps.GetStep(flannel.StepName),
 		steps.GetStep(kubelet.StepName),
 		steps.GetStep(cni.StepName),
```
stgleb: Why you add authorized keys twice?

gopherstein:
@stgleb @egopher We are reverting this, and it will need to be reopened... After applying this pull we had a number of etcd corruption issues. We tested the commit before this pull and the issue disappeared. We will need to do some further testing here.

PR #811, opened 2018-11-09
Add ability to list and proxy services

stgleb:
Add list services and self links gathered from kube-api
Add proxy for Supergiant that gets gets data from service
through kube-api proxy.

API

GET /v1/api/kubes/{kubename}/services

```
[
{"name":"alertmanager-operated",
 "type":"ClusterIP",
 "selfLink":"https://68.183.64.101/api/v1/namespaces/default/services/alertmanager-operated:9093/proxy"
}
]

```

Get actual service output 
TODO(stgleb): Change post to get and endode url in query param

GET /v1/api/kubes/{kubename}/services?url=https://68.183.64.101/api/v1/namespaces/default/services/alertmanager-operated:9093/proxy


pkg/kube/handler.go
```diff
@@ -837,3 +883,115 @@ func (h *Handler) getNodesMetrics(w http.ResponseWriter, r *http.Request) {
...
+
+	for key := range k.Masters {
+		if k.Masters[key] != nil {
+			masterNode = k.Masters[key]
+		}
+	}
+
+	url := fmt.Sprintf("https://%s/%s", masterNode.PublicIp, servicesUrl)
+	req, err := http.NewRequest(http.MethodGet, url, nil)
+	req.SetBasicAuth("root", "1234")
```
egopher: Why do we need that? Aren't we're using x509 authorization mode?

gopherstein:
@josephkane is testing.

pkg/kube/handler.go
stgleb: I've tried to disable basic auth, but without it kube-api server doesn't work for me, this needs to be addressed, but so far this is workaround to move forward. @egopher

pkg/kube/handler.go
```diff
@@ -36,6 +36,51 @@ type nodeProvisioner interface {
 	ProvisionNodes(context.Context, []profile.NodeProfile, *model.Kube, *steps.Config) ([]string, error)
 }
 
+type K8SServices struct {
+	Kind       string `json:"kind"`
```
saheienko: You could use `k8s.io/api/core/v1.Service` instead of custom one.

PR #833, opened 2018-11-10
Cancel Provisioning

josephkane:
* change communication when deleting a cluster that is provisioning


josephkane:
fix for https://github.com/supergiant/supergiant/issues/817

PR #838, opened 2018-11-12
Update certificates template to use publicIP from SG

stgleb:


templates/certificates.tpl
```diff
@@ -3,7 +3,7 @@ KUBERNETES_SSL_DIR={{ .KubernetesConfigDir }}/ssl
 sudo mkdir -p ${KUBERNETES_SSL_DIR}
 
 #sleep 10
-#FLANNELIP=$(ifconfig flannel.1|grep 'inet addr'| awk '{print substr($2,6,12)}')
+FLANNELIP=$(ifconfig flannel.1|grep 'inet addr'| awk '{print substr($2,6,12)}')
```
egopher: What if there is no flannel.1 interface?

pkg/workflows/steps/certificates/certificates.go
```diff
@@ -34,8 +35,12 @@ func New(script *template.Template) *Step {
 }
 
 func (s *Step) Run(ctx context.Context, out io.Writer, config *steps.Config) error {
-	config.CertificatesConfig.MasterHost = config.GetMaster().PrivateIp
-
+	if master := config.GetMaster(); master != nil {
```
saheienko: What master would be selected for the worker node?

templates/certificates.tpl
```diff
@@ -19,9 +19,10 @@ DNS.1 = kubernetes
 DNS.2 = kubernetes.default
 DNS.3 = kubernetes.default.svc
 DNS.4 = kubernetes.default.svc.cluster
-IP.1 = {MASTER_HOST}
-IP.2 = {PRIVATE_HOST}
-IP.3 = 10.3.0.1
+IP.1 = {FLANNELIP}
```
saheienko: How does it used?

templates/certificates.tpl
```diff
@@ -19,9 +19,10 @@ DNS.1 = kubernetes
...
+IP.2 = {MASTER_HOST}
+IP.3 = {PRIVATE_HOST}
```
saheienko: You can set addresses here without using sed.

pkg/workflows/steps/certificates/certificates.go
stgleb: Any active master node @saheienko

templates/certificates.tpl
stgleb: Removed that, that is obsolete

pkg/workflows/steps/certificates/certificates.go
stgleb: Either current node we are working on if it is master or any random active master https://github.com/supergiant/supergiant/blob/master/pkg/workflows/steps/config.go#L353

templates/certificates.tpl
stgleb: Removed

PR #893, opened 2018-11-15
Group Machines When Adding Node

josephkane:
* group machine types alphanumerically when adding nodes

* fixes https://github.com/supergiant/supergiant/issues/860

### DIFF
--- a/cmd/ui/assets/src/app/clusters/cluster/add-node/add-node.component.ts
+++ b/cmd/ui/assets/src/app/clusters/cluster/add-node/add-node.component.ts
@@ -127,7 +127,7 @@ export class AddNodeComponent implements OnInit, OnDestroy {
       this.providerSubj.pipe(
         filter(provider => provider === 'digitalocean'),
         switchMap(() => DOmachineSizes$),
-      ).subscribe(sizes => this.machineSizes$ = of(sizes))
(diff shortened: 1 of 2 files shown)


cmd/ui/assets/src/app/clusters/cluster/add-node/add-node.component.ts
```diff
@@ -127,15 +127,15 @@ export class AddNodeComponent implements OnInit, OnDestroy {
       this.providerSubj.pipe(
         filter(provider => provider === 'digitalocean'),
         switchMap(() => DOmachineSizes$),
-      ).subscribe(sizes => this.machineSizes$ = of(sizes))
```
stgleb: Just a quick questions, how do you sort items? lexicographically?

What  with case of

32gb
4gb

lexicographically 32gb is smaller than 4gb

josephkane: @stgleb great question. in speaking with @eldridgeh and @gopherstein, we decided that lexicographic sorting (basically grouping) is the scope of mvp. we don't have actual sizing data, so sorting 32gb vs s-12vcpu-48gb would take some serious time.

PR #920, opened 2018-11-17
App Store to Service Catalog

josephkane:
* rename App Store to Service Catalog

* change all related links

* fixes https://github.com/supergiant/supergiant/issues/827


josephkane:
also fixes https://github.com/supergiant/supergiant/issues/854

Issue #916, opened 2018-11-16
2.0: API - Ulimit on boxes needs to be set during provisioning.


eldridgeh:
**Short Summary:**
Ulimit on boxes need to be set on boxes while provisioning
**Steps to Reproduce:**
n/a
**Expected Results:**
n/a
**Actual Results:**
n/a
The ulimit for open files looks to be set at 1024, this needs to be increased to something high

stgleb:
@eldridgeh @saheienko Isn't this enough?

AWS

```
root@ip-172-31-13-132:~# ulimit -n
300000
```

DO

```
root@TestLimit-node-38aa:~# ulimit -n
300000
```

gopherstein:
Should be fine, just ensure it persists past reboot

stgleb:
it persists on AWS

```
cat /etc/security/limits.conf 
root soft  nofile 300000
root hard  nofile 300000
```
but in DO it doesn't

saheienko:
#1005 doesn't solve this issue.

stgleb:
Please, add steps to reproduce. I downsize ulimit on my local machine to 256 
but can't reproduce it.

stgleb:
Well, right after cluster is provisioned count of network connection constantly grows, profiling of goroutines shows leak 

```
goroutine profile: total 1808
896 @ 0x42fb5b 0x42ae99 0x42a546 0x497cea 0x497dfd 0x498b49 0x5bb79f 0x5cf298 0x62b1f9 0x62b6fd 0x62ec81 0x6e6465 0x516076 0x5161cf 0x6e6fb2 0x45d951
#	0x42a545	internal/poll.runtime_pollWait+0x65	/usr/local/go/src/runtime/netpoll.go:173
#	0x497ce9	internal/poll.(*pollDesc).wait+0x99	/usr/local/go/src/internal/poll/fd_poll_runtime.go:85
#	0x497dfc	internal/poll.(*pollDesc).waitRead+0x3c	/usr/local/go/src/internal/poll/fd_poll_runtime.go:90
#	0x498b48	internal/poll.(*FD).Read+0x178		/usr/local/go/src/internal/poll/fd_unix.go:169
#	0x5bb79e	net.(*netFD).Read+0x4e			/usr/local/go/src/net/fd_unix.go:202
#	0x5cf297	net.(*conn).Read+0x67			/usr/local/go/src/net/net.go:177
[...]

PR #963, opened 2018-11-21
Fix goroutine leak for helm proxy

saheienko:
Proxy portforwarder keeps connection to kube-apiserver open even if it is closed.

Related to https://github.com/supergiant/control/issues/916 (reduces number of connections)


gopherstein:
Which issue does this fix?

saheienko:
@gopherstein this is partially related to https://github.com/supergiant/control/issues/916.

@stgleb it could be no receivers for the error channel when protforwarder is closed.

stgleb:
Okay

PR #976, opened 2018-11-22
Handle errors while getting external IP gracefully

### LINKED ISSUE #975: AWS error whitelisting external IP
In rare cases supergiant receives an error

```
failed to whitelist supergiant IP in master security group: InvalidParameterValue: CIDR block /32 is malformed\n\tstatus code: 400
```

This is caused by error in figuring out external ip address and passing this error silently 

https://github.com/supergiant/control/blob/master/pkg/util/util.go#L143
https://github.com/supergiant/control/blob/master/pkg/util/util.go#L148

stgleb:
Closes #975

### DIFF
--- a/pkg/workflows/steps/amazon/util.go
+++ b/pkg/workflows/steps/amazon/util.go
@@ -1,0 +1,61 @@
+package amazon
+
+import (
+	"context"
+	"time"
+	"net/http"
+	"io/ioutil"
+	"strings"
+
+	"github.com/sirupsen/logrus"
+)
+
+func FindOutboundIP(ctx context.Context, findExternalIP func() (string, error)) (string, error) {
+	var (
+		timeout  = time.Second * 10
+		attempts = 5
+		publicIP string
+		err      error
+	)
+
+	for i := 0; i < attempts; i++ {
+		select {
+		case <-ctx.Done():
+			return "", ctx.Err()
+		default:
+			publicIP, err = findExternalIP()
+
+			if err == nil {
+				return publicIP, nil
+			} else {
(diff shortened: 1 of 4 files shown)


pkg/workflows/steps/amazon/util.go
```diff
@@ -0,0 +1,61 @@
...
+				timeout = timeout * 2
+			}
+		}
+	}
+
+	return publicIP, err
+}
+
+func findOutBoundIP() (string, error) {
+	serviceURL := "http://myexternalip.com/raw"
```
saheienko: In case of `http://myexternalip.com/raw` outage cluster bootstrapping would be failed?

stgleb: @saheienko yes, do you know better ways to figure out real external IP addr in network with random topology?

TheKLARKEN: You can curl the AWS IP:

`curl http://169.254.169.254/latest/meta-data/public-ipv4`
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-instance-addressing.html

TheKLARKEN: @stgleb ^ Would that work?

stgleb: @TheKLARKEN this is for instance public IP address, but I need to get public IP of SG itself which
can be outside of AWS.

TheKLARKEN: @stgleb Ahh, of SG itself. That makes sense. 🙂

Issue #939, opened 2018-11-20
2.0: UI - After installing an app, the UI should redirect to that clusters app section on the details page.


gopherstein:
![Screen Shot 2018-11-20 at 8:29 PM](http://marker.screenshots.prod.s3.amazonaws.com/51c0bba6096949f8835c49205d1499fe-1542745747059.png)
---
### More Info

**Reported by:** eldridge henley (eldridge@qbox.io)
**Source URL**: [http://localhost:4200/catalog/supergiant/details/supergiant](http://localhost:4200/catalog/supergiant/details/supergiant)

<table><tr><td><strong>Browser</strong></td><td>Chrome 70.0.3538.77</td></tr><tr><td><strong>Screen Size</strong></td><td>1680 x 1050</td></tr><tr><td><strong>OS</strong></td><td>OS X 10.14.1</td></tr><tr><td><strong>Viewport Size</strong></td><td>1623 x 1027</td></tr><tr><td><strong>Zoom Level</strong></td><td>100%</td></tr><tr><td><strong>Pixel Ratio</strong></td><td>@2x</td></tr></table>

Issue #931, opened 2018-11-20
2.0: UI - Fix table view on hover 


hetmanchuk:
![image](https://user-images.githubusercontent.com/11389252/48775841-e55c7b80-ecd6-11e8-8ebb-2162e12d6915.png)

**Steps to Reproduce:**
1. Spin up minimum 1 cluster
2. Go to Dashboard
3. Switch to "Table view" on the right top corner

@josephkane

PR #983, opened 2018-11-26
Route to Cluster After Deploy

josephkane:
* route to target cluster after deploying app

* fix hover state on dashboard cluster table

* fixes https://github.com/supergiant/control/issues/939, https://github.com/supergiant/control/issues/931

### DIFF
--- a/cmd/ui/assets/src/app/features/app-store/app-details/deploy/deploy.component.ts
+++ b/cmd/ui/assets/src/app/features/app-store/app-details/deploy/deploy.component.ts
@@ -71,7 +71,7 @@ export class DeployComponent implements OnInit {
       if (result instanceof ErrorEvent) {
         return;
       }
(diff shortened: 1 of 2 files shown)


cmd/ui/assets/src/app/features/app-store/app-details/deploy/deploy.component.ts
```diff
@@ -71,7 +71,7 @@ export class DeployComponent implements OnInit {
       if (result instanceof ErrorEvent) {
         return;
       }
-      this.router.navigate(['apps']);
```
stgleb: I may be wrong, but don't we  name is as kubes?

josephkane: @stgleb for the api routes, we use 'kubes', for urls in the ui we use 'clusters'

PR #1009, opened 2018-11-27
Fix prometheus proxy in kube-api proxy

stgleb:
To work properly Prometheus should be awared about proxy it works behind


stgleb:
How to check

http://{{ ,MasterIP }}:8080/api/v1/namespaces/default/services/prometheus-operated:9090/proxy/graph
http://{{ .MasterIP }}:8080/api/v1/namespaces/default/services/alertmanager-operated:9093/proxy/#/alerts

these links must work correctly

PR #1017, opened 2018-11-27
fix (add-node.component) change account type source property

nikage:


nikage:
Updated PR, this should work

PR #906, opened 2018-11-16
ETCD Static configration

egopher:
https://github.com/supergiant/supergiant/issues/895
https://github.com/supergiant/supergiant/issues/864

PR #1024, opened 2018-11-27
Allow traffic between master nodes in AWS

### LINKED ISSUE #1022: AWS multi master networking is broken
**Short Summary:**

Quote 

```
Instances in the same security group will not be able to connect to each other out of the box
```
Master nodes in cluster needs to communicate - etcd  management port.
So AWS multi master cluster will be broken because nodes of etcd don't see
each other and never gets to consensus.

**Steps to Reproduce:**

1. Create cluster in AWS with three nodes
2. ssh to one node and try to ping, curl or whatever another node

**Expected Results:**

Ping, ssh, curl will work fine

**Actual Results:**

It doesn't work

Reproduces on current master

stgleb:
By default AWS forbids traffic even between nodes in the same security group
This works for worker nodes which do not communicate between each other,
but it doesn't work for master nodes whose etcd instances needs send heartbeats
all the time. 

Closes #1022


gopherstein:
closing as it is a dup fix with https://github.com/supergiant/control/pull/906

PR #1030, opened 2018-11-28
task/S20-442-proxy-backend

taaraora:
Proxy is implemented.
Default proxy port range is 60200 - 60250


pkg/kube/handler.go
```diff
@@ -979,6 +986,7 @@ func (h *Handler) getNodesMetrics(w http.ResponseWriter, r *http.Request) {
 }
 
 func (h *Handler) getServices(w http.ResponseWriter, r *http.Request) {
+	fmt.Println("OLOLO")
```
egopher: please remove it.

pkg/kube/proxy.go
```diff
@@ -0,0 +1,172 @@
...
+func (p *APIProxy) SetServices(kubeID string, services []*ServiceInfo) error {
+	if p.svc == nil || p.k8sClusterIDToProxies == nil {
+		return errors.New("component was not initialized properly")
+	}
+
+	p.servicesMux.Lock()
+
+	_, exists := p.k8sClusterIDToProxies[kubeID]
+	if !exists {
+		p.k8sClusterIDToProxies[kubeID] = make(map[string]*ServiceProxy)
```
egopher: This will cause race conditions in case of concurrent requests. Please either add a mutex there or use sync.Map.

cmd/controlplane/main.go
```diff
@@ -8,20 +8,24 @@ import (
...
-	uiDir         = flag.String("ui-dir", "./cmd/ui/assets/dist", "directory for supergiant ui to be served")
+	addr                 = flag.String("address", "0.0.0.0", "network interface to attach server to")
+	port                 = flag.Int("port", 8080, "tcp port to listen for incoming requests")
+	etcdURL              = flag.String("etcd-url", "localhost:2379", "etcd url with port")
+	templatesDir         = flag.String("templates", "/etc/supergiant/templates/", "supergiant will load script templates from the specified directory on start")
+	logLevel             = flag.String("log-level", "INFO", "logging level, e.g. info, warning, debug, error, fatal")
+	logFormat            = flag.String("log-format", "txt", "logging format [txt json]")
+	spawnInterval        = flag.Int("spawnInterval", 5, "interval between API calls to cloud provider for creating instance")
+	uiDir                = flag.String("ui-dir", "./cmd/ui/assets/dist", "directory for supergiant ui to be served")
+	ProxiesPortRangeFrom = flag.Int("proxies-port-from", 60200, "first tcp port in a range of binding reverse proxies for service apps")
```
saheienko: Please, use a single flag for setting a port range.

example: https://github.com/kubernetes/kubernetes/blob/master/cmd/kube-apiserver/app/options/options.go#L185

pkg/kube/handler.go
```diff
@@ -1003,12 +1009,12 @@ func (h *Handler) getServices(w http.ResponseWriter, r *http.Request) {
...
 		kubeID      string
 		masterNode  *node.Node
-		k8sServices = &K8SServices{}
+		k8sServices = &k8SServices{}
 	)
 	vars := mux.Vars(r)
 	kubeID = vars["kubeID"]
 
-	k, err := h.svc.Get(r.Context(), kubeID)
+	k8sCluster, err := h.svc.Get(r.Context(), kubeID)
```
saheienko: Please, use short names for local variables.
https://github.com/golang/go/wiki/CodeReviewComments#variable-names

pkg/kube/handler.go
```diff
@@ -1057,122 +1063,85 @@ func (h *Handler) getServices(w http.ResponseWriter, r *http.Request) {
...
+				ID:        service.Metadata.UID,
+				Name:      service.Metadata.Name,
+				Type:      service.Spec.Type,
+				Namespace: service.Metadata.Namespace,
+			}
+			serviceInfos = append(serviceInfos, serviceInfo)
+
+			targetServices = append(targetServices, &proxy.Target{
+				kubeID + service.Metadata.UID,
+				fmt.Sprintf("https://%s%s:%d/proxy",
```
saheienko: Please, set fields explicitly.
```
struct{f1: v1, f2: v2} instead of struct{v1, v2} 
```

pkg/kube/handler.go
```diff
@@ -1057,122 +1063,85 @@ func (h *Handler) getServices(w http.ResponseWriter, r *http.Request) {
...
+						ID:        service.Metadata.UID,
+						Name:      service.Metadata.Name,
+						Type:      service.Spec.Type,
+						Namespace: service.Metadata.Namespace,
 					}
-					services = append(services, service)
+					serviceInfos = append(serviceInfos, serviceInfo)
+
+					targetServices = append(targetServices, &proxy.Target{
+						kubeID + service.Metadata.UID,
```
saheienko: The unique identifier for services is namespace and name.

`UID` is the unique in time and space value for this object(if service is updated it will get another id).
https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#uids

pkg/kube/handler.go
```diff
@@ -1057,122 +1063,85 @@ func (h *Handler) getServices(w http.ResponseWriter, r *http.Request) {
...
-	vars := mux.Vars(r)
-	kubeID := vars["kubeID"]
-
-	k, err := h.svc.Get(r.Context(), kubeID)
-	if err != nil {
-		if sgerrors.IsNotFound(err) {
-			message.SendNotFound(w, kubeID, err)
-			return
+func labelIsTrue(labelName string, labels map[string]string) bool {
+	for labelKey, labelValue := range labels {
```
saheienko: No need to iterate over a map:
```
v, ok := labels[name]
```
Also, could it more generic, like `contains(name, value, labels)`?

pkg/proxy/proxy.go
```diff
@@ -0,0 +1,225 @@
...
+				logger.Errorf("error while serving address: %v", listener.Addr())
+			}
+		}
+	}()
+
+	logger.Infof("proxy server started on: %s, for targetURL: %+v", addr, targetURL)
+
+	return &ServiceReverseProxy{
+		srv:         httpServer,
+		servingBase: addr,
```
saheienko: Does `serviceBase` is a good name for server address?

pkg/proxy/proxy.go
```diff
@@ -0,0 +1,225 @@
...
+	}, nil
+}
+
+func (p *ReverseProxyContainer) RegisterProxies(targets []*Target) error {
+	if p.Proxies == nil {
+		return errors.New("component was not initialized properly")
```
saheienko: Please, add component name to the error message.

pkg/proxy/proxy.go
```diff
@@ -0,0 +1,225 @@
...
+		if strings.HasPrefix(inputURL, selfLink) {
+			inputURL = strings.TrimPrefix(inputURL, selfLink)
+		}
+
+		if strings.Index(inputURL, ":") == 0 {
+			parts := strings.Split(inputURL, "/")
+			inputURL = strings.TrimPrefix(inputURL, parts[0]+"/proxy")
+		}
+
+		logger.Infof("req.URL: %+v, inputURL %s, inputFullURL: %s",
```
saheienko: It's a kind of a spamming message, use `debug` logging level.

pkg/proxy/proxy.go
```diff
@@ -0,0 +1,225 @@
...
+		// Update the headers to allow for SSL redirection
+		req.Header.Set("X-Forwarded-Host", req.Header.Get("Host"))
+		req.URL.Path = inputURL
+
+		// Note that ServeHttp is non blocking and uses a go routine under the hood
+		reverseProxy.ServeHTTP(res, req)
+	}
+}
+
+func (p *ReverseProxyContainer) Shutdown(ctx context.Context) {
```
saheienko: Is is possible to shut down some particular proxy?

templates/prometheus.sh.tpl
```diff
@@ -2,8 +2,6 @@
 # it explicitly
 sudo /opt/bin/helm install stable/prometheus-operator \
     --name=prometheus-operator \
-    --set prometheus.prometheusSpec.externalUrl=/api/v1/namespaces/default/services/prometheus-operated:9090/proxy/ \
```
saheienko: How does it related to service proxy?
Please, make a separate PR if it doesn't related.

pkg/kube/handler.go
taaraora: Sorry, that is not my scenario.

pkg/kube/handler.go
taaraora: done

pkg/kube/handler.go
taaraora: I didn't get your point.

pkg/kube/handler.go
taaraora: Oh, nice catch, thank you!

pkg/kube/handler.go
saheienko: Don't use `service.Metadata.UID` as unique service identifier.

cmd/controlplane/main.go
```diff
@@ -8,20 +8,24 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/supergiant/control/pkg/proxy"
```
egopher: gofmt

pkg/controlplane/server.go
```diff
@@ -4,9 +4,13 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"net/url"
 	"os"
+	"strings"
 	"time"
 
+	"github.com/supergiant/control/pkg/proxy"
```
egopher: remove whitespace and gofmt

templates/prometheus.sh.tpl
egopher: It breaks service proxy.

PR #1035, opened 2018-11-28
Add StaticAuth cluster credentials to profiles

saheienko:
With this changes it would be possible to provide static password or token credentials for authentication during bootstrap process.

This should be set to profile model:
```
{
...
  "staticAuth": {
    "basicAuth": {
      "password": "userPassword",
      "name": "userName",
      "id":"userId",
      "groups": ["group1", "group2"]
    },
    "tokens": {
      "token": "userToken",
      "name": "userName",
      "id":"userId",
      "groups": ["group1", "group2"]
    }
  }
...
}
```

### DIFF
--- a/templates/certificates.tpl
+++ b/templates/certificates.tpl
@@ -51,15 +51,13 @@ sudo chmod 600 /etc/kubernetes/ssl/*-key.pem
 sudo chown root:root /etc/kubernetes/ssl/*-key.pem
 
 sudo bash -c "cat > /etc/kubernetes/ssl/basic_auth.csv <<EOF
-{{ .Password }},{{ .Username }},admin
+{{- range .StaticAuth.BasicAuth }}
+{{ .Password }},{{ .Name }},{{ .ID }},{{ stringsJoin .Groups "," }}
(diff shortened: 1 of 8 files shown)


templates/certificates.tpl
```diff
@@ -51,15 +51,13 @@ sudo chmod 600 /etc/kubernetes/ssl/*-key.pem
 sudo chown root:root /etc/kubernetes/ssl/*-key.pem
 
 sudo bash -c "cat > /etc/kubernetes/ssl/basic_auth.csv <<EOF
-{{ .Password }},{{ .Username }},admin
+{{- range .StaticAuth.BasicAuth }}
+{{ .Password }},{{ .Name }},{{ .ID }},{{ stringsJoin .Groups "," }}
+{{ end -}}
 EOF"
```
stgleb: Will it be ok for worker to authorize itself?

https://github.com/supergiant/control/blob/master/templates/manifest.sh.tpl#L203

saheienko: Yes, it will. This user has added to basic auth credentials.

PR #1067, opened 2018-11-29
Clean-up resources AWS

### LINKED ISSUE #1052: 2.0: API - Resource pollution
When I delete cluster resources like VPC, keys, subnets, security groups are left, 
they need to be deleted with cluster.

stgleb:
Delete Security Groups

Partially closes #1052


pkg/controlplane/server.go
```diff
@@ -245,8 +245,10 @@ func configureApplication(cfg *Config) (*mux.Router, error) {
 	amazon.InitCreateSecurityGroups(amazon.GetEC2)
 	amazon.InitCreateVPC(amazon.GetEC2)
 	amazon.InitCreateSubnet(amazon.GetEC2)
-	amazon.InitDeleteCluster(amazon.GetEC2)
+	amazon.InitDeleteClusterMachines(amazon.GetEC2)
 	amazon.InitDeleteNode(amazon.GetEC2)
+	amazon.InitDeleteSecurityGroup(amazon.GetEC2)
+	amazon.InitDeleteVPC(amazon.GetEC2)
```
saheienko: Could it be a part of deleteCluster step?

pkg/controlplane/server.go
```diff
@@ -245,8 +245,10 @@ func configureApplication(cfg *Config) (*mux.Router, error) {
 	amazon.InitCreateSecurityGroups(amazon.GetEC2)
 	amazon.InitCreateVPC(amazon.GetEC2)
 	amazon.InitCreateSubnet(amazon.GetEC2)
```
saheienko: There is no DeleteSubnetes step.

pkg/controlplane/server.go
stgleb: Well, we dont have delete cluster step anymore. We have delete cluster workflow which includes

Deletion of instances
Deletion of security groups

TBD

Deletion of subntes 
Deletion of VPC

pkg/controlplane/server.go
stgleb: We have it, but I didn't initialize it since currently creation of subnets doesn't work. We use default subnet in default VPC. So currently I delete those entities that were really created in provisioner - security groups and instances.

Issue #957, opened 2018-11-20
2.0: UI - User should be provided with info when an app is installed.


eldridgeh:
When you deploy an app to a cluster, the user should be provided with the helm install return. 

The user should also be able to expand details for the app from the cluster details page and show app feedback, details, ports, etc.

gopherstein:
Just like after cluster provisioning, the user needs an after app install call out.

gopherstein:
Adding @hetmanchuk as the first responder on this, as it is more of a user flow issue. Once we have a plan we can re-assign to the frontend.

gopherstein:
User testing  #1, #2 feedback: Need feedback about how to access my app.

gopherstein:
This is a standard bit of data returned by helm after install. We just need to be displaying it to the user and saving it in a details tab for the deployed app.

gopherstein:
User testing #2 feedback: 

* "I'm at the point where I'm like, I'm not sure where to go from here."
    * "I have a Docker image we would deploy to production. From here, I'm not sure if it's expected for users to leave UI and do that the way they traditionally would."

josephkane:
@saheienko per war room discussion, we will need a GET endpoint for a release that returns release details and NOTES.txt

gopherstein:
Looks like the backend is merged.

PR #1073, opened 2018-11-30
Release Info

josephkane:
* allow users to acces information about deployed apps from cluster details page

* fix for https://github.com/supergiant/control/issues/957

### DIFF
--- a/cmd/ui/assets/src/app/app.module.ts
+++ b/cmd/ui/assets/src/app/app.module.ts
@@ -155,6 +155,7 @@ import { DeleteClusterModalComponent } from './clusters/cluster/delete-cluster-m
 import { DeleteReleaseModalComponent } from './clusters/cluster/delete-release-modal/delete-release-modal.component';
 import { SshCommandsModalComponent } from './clusters/cluster/ssh-commands-modal/ssh-commands-modal.component';
 import { KubectlConfigModalComponent } from './clusters/cluster/kubectl-config-modal/kubectl-config-modal.component';
+import { ReleaseInfoModalComponent } from './clusters/cluster/release-info-modal/release-info-modal.component';
 
 @NgModule({
   declarations: [
@@ -213,6 +214,7 @@ import { KubectlConfigModalComponent } from './clusters/cluster/kubectl-config-m
     DeleteReleaseModalComponent,
     SshCommandsModalComponent,
     KubectlConfigModalComponent,
+    ReleaseInfoModalComponent,
   ],
   imports: [
     BrowserModule,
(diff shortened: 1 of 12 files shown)


cmd/ui/assets/src/app/app.module.ts
```diff
@@ -317,7 +319,8 @@ import { KubectlConfigModalComponent } from './clusters/cluster/kubectl-config-m
     DeleteClusterModalComponent,
     DeleteReleaseModalComponent,
     SshCommandsModalComponent,
-    KubectlConfigModalComponent
+    KubectlConfigModalComponent,
+    ReleaseInfoModalComponent
```
nikage: why the components are included into import section?

cmd/ui/assets/src/app/clusters/cluster/cluster.component.scss
```diff
@@ -755,6 +755,20 @@
     }
   }
 
+  .info-icon {
+    background-image: url(/assets/img/info-pink.svg);
```
nikage: this won't work with aot build, I think inital slash should be removed. needs to be checked out with `npm run build:prod`

cmd/ui/assets/src/app/app.module.ts
josephkane: modals are included here so they are loaded dynamically

cmd/ui/assets/src/app/clusters/cluster/cluster.component.scss
josephkane: thanks, will check and fix file-wide if it is an issue

josephkane: @nikage I wasn't able to get this to break, it seems to work fine. would you mind fetching this PR and testing it?

cmd/ui/assets/src/app/app.module.ts
nikage: yeap, sorry. Github line collapsing confused me. Then that is fine

cmd/ui/assets/src/app/clusters/cluster/cluster.component.scss
nikage: this won't break the build process itself, but the image link might be broken. I'll check this in a while

josephkane: thats what I mean, I couldn't get the image link to break. I might be doing something wrong though

eldridgeh:
This looks to be working. However, was it intended to split the cluster name on the inline machine info? This probably isn't a big deal, but I don't think we do this anywhere else anymore
screen shot 2018-11-30 at 11 58 05 am
.

cmd/ui/assets/src/app/clusters/cluster/cluster.component.scss
nikage: yeap there is an issue. Here is the icon on localhost:4200 e.g dev server
image

nikage: and here is localhost:8080 e.g. build 
image

nikage: apparently, no. it will be okay.  that was a browser cache issue, not the code

PR #1090, opened 2018-12-01
Fix task restart

stgleb:
SG replies with 500 error when task 
has been failed on create instance step 
and didn't obtain public IP

### DIFF
--- a/pkg/workflows/handler.go
+++ b/pkg/workflows/handler.go
@@ -158,7 +158,7 @@ func (h *TaskHandler) RestartTask(w http.ResponseWriter, r *http.Request) {
 	task, err := DeserializeTask(data, h.repository)
 
 	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/workflows/handler.go
```diff
@@ -146,16 +146,19 @@ func (h *TaskHandler) RestartTask(w http.ResponseWriter, r *http.Request) {
...
-		http.Error(w, err.Error(), http.StatusInternalServerError)
+		logrus.Debugf("task %s not found", id)
+		http.NotFound(w, r)
 		return
 	}
 
 	task, err := DeserializeTask(data, h.repository)
 
 	if err != nil {
+		logrus.Debugf("error deserializing task %v", err)
```
saheienko: Please add task id/name here. Does it a debug log message?

stgleb: Done

PR #1096, opened 2018-12-03
Add az for aws on machine config step cluster creation

nikage:


cmd/ui/assets/src/app/clusters/new-cluster/new-cluster.component.ts
```diff
@@ -282,12 +295,14 @@ export class NewClusterComponent implements OnInit, OnDestroy {
 
   validMachine(machine) {
     if (
-        machine.machineType != null &&
-        machine.role != null &&
-        typeof(machine.qty) == "number"
-      ) {
+      machine.machineType != null &&
+      machine.role != null &&
+      typeof(machine.qty) == "number"
```
josephkane: we should validate that all aws machines have an availability zone here

PR #1097, opened 2018-12-03
Fix connection leak

### LINKED ISSUE #916: 2.0: API - Ulimit on boxes needs to be set during provisioning.

stgleb:
Reuse client and transport  objects to prevent connection leak between SG and kube-api
Bu,p k8s version to include https://github.com/kubernetes/kubernetes/pull/68041

Closes #916

### DIFF
--- a/pkg/controlplane/server.go
+++ b/pkg/controlplane/server.go
@@ -8,7 +8,6 @@ import (
 	"os"
 	"strings"
 	"time"
-	_ "net/http/pprof"
 
 	"k8s.io/helm/pkg/repo"
 
@@ -287,11 +286,6 @@ func configureApplication(cfg *Config) (*mux.Router, error) {
 		TokenService: jwtService,
 	}
 	protectedAPI.Use(authMiddleware.AuthMiddleware, api.ContentTypeJSON)
-
-	// Run pprof http server
-	go func(){
-		logrus.Info(http.ListenAndServe("localhost:6060", nil))
-	}()
 
 	if err := serveUI(cfg, router); err != nil {
 		return nil, err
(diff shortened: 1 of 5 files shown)


pkg/controlplane/server.go
```diff
@@ -286,6 +288,11 @@ func configureApplication(cfg *Config) (*mux.Router, error) {
 	}
 	protectedAPI.Use(authMiddleware.AuthMiddleware, api.ContentTypeJSON)
 
+	// Run pprof http server
+	go func(){
+		logrus.Info(http.ListenAndServe("localhost:6060", nil))
```
saheienko: Please, open a separate PR for this feature.
Also, it shouldn't be enabled by default.

egopher: Can it be optional with a flag?

pkg/kube/handler.go
```diff
@@ -916,7 +986,7 @@ func (h *Handler) getClusterMetrics(w http.ResponseWriter, r *http.Request) {
 
 	for metricType, relUrl := range metricsRelUrls {
 		url := fmt.Sprintf("https://%s/%s/%s", masterNode.PublicIp, baseUrl, relUrl)
-		metricResponse, err := h.getMetrics(url)
+		metricResponse, err := h.getMetrics(url, k)
```
saheienko: Could you use kubernetes client for retrieving these metrics? Kubernetes client solves this issues.

stgleb: Will do in separate PR, just want to fix it as fast as possible.

PR #1099, opened 2018-12-03
Create instance IAM profiles for AWS

saheienko:
Create instance profiles for mater/node roles on AWS.
Use policies from https://github.com/kubernetes/cloud-provider-aws#iam-policy


pkg/workflows/steps/amazon/create_machine.go
```diff
@@ -68,18 +71,30 @@ func (s *StepCreateInstance) Run(ctx context.Context, w io.Writer, cfg *steps.Co
 		secGroupID = &cfg.AWSConfig.NodesSecurityGroupID
 	}
 
+	// TODO: reuse sessions
```
stgleb: This is done for each machine before it is created? Maybe move it to separate step?

saheienko: Yes, this ensures that instance profile exists before it will be run.
It could be moved to a separate step too.

stgleb: Ok, please add todo for that. Don't want to forget it

stgleb:
Please resolve the conflict and I'll merge it

PR #1100, opened 2018-12-03
Add pprof handlers

stgleb:


pkg/controlplane/server.go
```diff
@@ -290,6 +295,13 @@ func configureApplication(cfg *Config) (*mux.Router, error) {
 	}
 	protectedAPI.Use(authMiddleware.AuthMiddleware, api.ContentTypeJSON)
 
+	if cfg.PprofListenStr != "" {
+		go func() {
+			logrus.Debugf("Start pprof on %s", cfg.PprofListenStr)
+			logrus.Info(http.ListenAndServe(cfg.PprofListenStr, nil))
```
saheienko: Could it be added to the Control mux?
example: 
https://github.com/kubernetes/kubernetes/blob/f48e18faa4dc035cc927c6a2b34c83c8475b55fa/cmd/controller-manager/app/serve.go#L58
https://github.com/kubernetes/apiserver/blob/master/pkg/server/routes/profiling.go#L30

stgleb: I would like to leave it on separate server, because `go tool` doesn't support authentication and we will expose pprof endpoints to outside world without any authentication. This can be potential attack vector if someone runs go tool against SG server.

Issue #1006, opened 2018-11-27
2.0: API - Prometheus needs to be deployed under kube-system 


eldridgeh:
**Short Summary:**
Prometheus needs to be deployed under the kube-system namespace.

PR #1108, opened 2018-12-03
[WIP] Made prometheus to be installed to the kube-system by default.

egopher:


gopherstein:
fixes https://github.com/supergiant/control/issues/1006

gopherstein:
@egopher This seems to break @stgleb 's metric proxy. I assume his proxy is looking for the default namespace.

gopherstein:
Also, is there a way we can add a label to a service through helm?

The grafana service that gets created during install needs to have the `kubernetes.io/cluster-service: "true"` label applied to it.

gopherstein:
example: https://github.com/supergiant/charts/blob/master/traefik/templates/traefik-service-ui.yaml

PR #1120, opened 2018-12-04
Add machine types search on adding node to cluster

nikage:


gopherstein:
On AWS you need to hit enter for the selected machine type to be selected, you cant click to choose it with the mouse.

PR #1130, opened 2018-12-04
Made provider string to be passed only in AWS

egopher:


pkg/workflows/steps/manifest/manifest.go
```diff
@@ -49,6 +50,10 @@ func (j *Step) Run(ctx context.Context, out io.Writer, config *steps.Config) err
 		}
 	}
 
+	if config.Provider != clouds.AWS {
+		config.ManifestConfig.ProviderString = ""
```
stgleb: maybe it is better to change it here https://github.com/supergiant/control/blob/master/pkg/workflows/steps/config.go#L480 ?

saheienko:
Cloud provider is setting only for AWS and GCE. 
https://github.com/supergiant/control/blob/master/pkg/workflows/steps/config.go#L476
https://github.com/supergiant/control/blob/master/pkg/workflows/steps/config.go#L476

Do you have any issues with it?

stgleb:
@saheienko Yes, DO provisioning doesn't work after this PR. I gues the problem can be here
https://github.com/supergiant/control/blob/master/templates/kubelet.tpl#L35

You leave a blank line and downside params are omitted

Issue #1137, opened 2018-12-05
AWS resource management


stgleb:
To be able to clean up resources allocated for cluster we need to have a control over it.
Usage of default vpcs, subnets, route tables, route table associations etc. leads to error during deletion and more over
it can cause erros during creation of a cluster.

Resolution:
Cluster need to allocate all resources by itself and clean them up.

PR #1138, opened 2018-12-05
Make AWS work again

### LINKED ISSUE #1052: 2.0: API - Resource pollution

stgleb:
Create all resources need for aws cluster by SG

Flow

1. VPC
2. Security Groups
3. KeyPair
4. Instance Profiles
5. Create and attach Internet Gateway to VPC
6. Subnets per each AZ in region
7. Route table
8. Associate route table with subnet

From API point of view profile now has additional parameter

Subnets - this is a map of availability zone to subnet.

For the sake of reliability I strongly recommend disable use of existing subnet.

Closes #1052 #1137


gopherstein:
@stgleb Fix conflicts. Then we are good to merge. Testing went well.

PR #1143, opened 2018-12-05
Google Compute Engine UI/Backend

stgleb:


pkg/provisioner/util.go
```diff
@@ -21,6 +21,7 @@ import (
 	"github.com/supergiant/control/pkg/util"
 	"github.com/supergiant/control/pkg/workflows"
 	"github.com/supergiant/control/pkg/workflows/steps"
+	"strings"
 )
 
 type EtcdTokenGetter struct {
```
egopher: We don't need this anymore. Current implementation is doesn't use public etcd bootstrap tokens and set's up a static configuration for the etcd cluster.

stgleb: Done

stgleb:
@eldridgeh Please check it again against AWS in this branch. Please don't forget  rebuil UI and bypass browser cache.

Issue #1103, opened 2018-12-03
2.0.1: UI - Adding a node should show Tasks and Logs for the new node


gopherstein:
![Screen Shot 2018-12-03 at 4:18 PM](http://marker.screenshots.prod.s3.amazonaws.com/5730eb16f0b64d7ba08409c2839566c9-1543853935781.png)

**Summary:**
When adding a node to an existing cluster, we currently do not display any of the tasks or logs while the new node is provisioning. 
**Steps to Reproduce:** 
1. Spin up a cluster
2. Add a node
**Expected Results:** 
When adding a node, the user is redirected to the Cluster details page at the node level.
The user is able to see in the machines section the ongoing tasks and logs as the node is provisioned. Just like they do when creating the cluster.
**Actual Results:** 
A node is added and the user does not see any task or logs.
**Notes:**
Adding a node currently returns a 200, it needs to return a 202 instead.
If you have any questions, please ask.

---
### More Info

**Reported by:** eldridge henley (eldridge@qbox.io)
**Source URL**: [http://localhost:4200/clusters/7dd65dfc](http://localhost:4200/clusters/7dd65dfc)

<table><tr><td><strong>Browser</strong></td><td>Chrome 70.0.3538.110</td></tr><tr><td><strong>Screen Size</strong></td><td>1680 x 1050</td></tr><tr><td><strong>OS</strong></td><td>OS X 10.14.1</td></tr><tr><td><strong>Viewport Size</strong></td><td>1680 x 1020</td></tr><tr><td><strong>Zoom Level</strong></td><td>100%</td></tr><tr><td><strong>Pixel Ratio</strong></td><td>@2x</td></tr></table>

nikage:
@gopherstein  @hetmanchuk is there additional design supposed to this feature?

josephkane:
@hetmanchuk 
![image](https://user-images.githubusercontent.com/18269936/49442791-2ac98000-f790-11e8-81bf-c4b939b29876.png)

PR #1145, opened 2018-12-05
New Node Progress

josephkane:
* show task logs and steps for nodes added to a cluster

* fix for https://github.com/supergiant/control/issues/1103

### DIFF
--- a/cmd/ui/assets/src/app/clusters/cluster/cluster.component.ts
+++ b/cmd/ui/assets/src/app/clusters/cluster/cluster.component.ts
@@ -42,8 +42,11 @@ export class ClusterComponent implements OnInit, OnDestroy {
   public kubeString: string;
 
   // machine list vars
-  machines: any;
-  machineListColumns = ["state", "role", "size", "name", "cpu", "ram", "region", "publicIp", "delete"];
+  activeMachines: any;
+  activeMachineListColumns = ["state", "role", "size", "name", "cpu", "ram", "region", "publicIp", "delete"];
+
+  nonActiveMachines: any;
+  nonActiveMachineListColumns = ["state", "role", "size", "name", "region", "steps", "logs", "delete"];
 
   // task list vars
   tasks: any;
@@ -117,7 +120,6 @@ export class ClusterComponent implements OnInit, OnDestroy {
   }
 
   toggleSteps(task) {
-
     task.showSteps = !task.showSteps;
 
     if (this.expandedTaskIds.has(task.id)) {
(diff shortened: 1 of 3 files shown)


cmd/ui/assets/src/app/clusters/cluster/cluster.component.ts
```diff
@@ -269,9 +268,42 @@ export class ClusterComponent implements OnInit, OnDestroy {
...
+    this.activeMachines = new MatTableDataSource(activeMachines);
+    this.activeMachines.sort = this.sort;
+    this.activeMachines.paginator = this.paginator;
+
+    if (nonActiveMachines.length > 0) {
+      let executingTasksObj = {};
+
+      this.getKubeStatus(this.clusterId).subscribe(
+        tasks => {
+          let executing = tasks.filter(t => t.status == "executing");
```
nikage: this looks like non rxjs way(

cmd/ui/assets/src/app/clusters/cluster/cluster.component.ts
```diff
@@ -464,6 +496,7 @@ export class ClusterComponent implements OnInit, OnDestroy {
   }
 
   expandRow = (_, row) => row.hasOwnProperty('detailRow');
+  // expandRow = (_, row) => { console.log("WHEN: ", row); return row.hasOwnProperty('detailRow') };
```
nikage: pls remove commented code

josephkane: thanks, good catch

eldridgeh:
@josephkane 
1. Once the node is provisioned the provisioning section does not go away. I had to refresh the page. This happens in DO and AWS.
screen shot 2018-12-05 at 12 25 43 pm
2. It takes a few seconds for the provisioning section to show up when adding a node to a DO. Would it be possible to make it show up faster? This looked to be instant with AWS
3. I created a cluster in AWS and when it switched over to the cluster details page, the provisioning node table showed up showing that the only node in the cluster was provisioning. We should only see the provisioning table when adding a node. It did not do this with DO. 
4. This is more than likely related to 1, after deleting a node the provisioning table did not go away.

josephkane:
@eldridgeh 
1. great catch, fixed
2. This seem to be a symptom of A.) no "planned" for DO nodes or B.) DO node objects aren't created at the same time as AWS nodes are in the DB. This one is sort of out of my hands (I could be mistaken)
3. This is most likely due to the cluster being declared "operational" before the last node is finished provisioning. In that case, this table would work as designed (because that node actually is still provisioning)
4. fixed with 1

PR #1149, opened 2018-12-05
Default storage classes.

egopher:
Adding a default storage classes. For AWS and GCE it holds provider specific implementation, for others - it uses local by default. We should support more storage classes in the future and to provide a user an ability to choose implementation.


pkg/controlplane/server.go
```diff
@@ -3,6 +3,7 @@ package controlplane
 import (
 	"context"
 	"fmt"
```
stgleb: goimports

pkg/workflows/workflow.go
```diff
@@ -1,6 +1,7 @@
 package workflows
 
 import (
```
stgleb: goimports

pkg/workflows/steps/storageclass/storageclass_test.go
```diff
@@ -0,0 +1,28 @@
...
+func (f *fakeRunner) Run(command *runner.Command) error {
+	if len(f.errMsg) > 0 {
+		return errors.New(f.errMsg)
+	}
+
+	_, err := io.Copy(command.Out, strings.NewReader(command.Script))
+	return err
+}
+
+func TestStep_Run(t *testing.T) {
```
stgleb: Will you should not a test in this PR?

egopher: It's WIP PR, @stgleb

templates/storageclass.sh.tpl
```diff
@@ -0,0 +1,59 @@
...
+  type: gp2
+
+---
+
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: gp2
+  annotations:
+    storageclass.beta.kubernetes.io/is-default-class: \"true\"
```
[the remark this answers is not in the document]
egopher: :+1:

templates/storageclass.sh.tpl
```diff
@@ -0,0 +1,59 @@
...
+  labels:
+    k8s-addon: storage-aws.addons.k8s.io
+provisioner: kubernetes.io/aws-ebs
```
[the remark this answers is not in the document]
egopher: We have specific users feedback to have storage classes be set by default.

pkg/controlplane/server.go
```diff
@@ -50,7 +48,9 @@ import (
 	"github.com/supergiant/control/pkg/workflows/steps/poststart"
 	"github.com/supergiant/control/pkg/workflows/steps/prometheus"
 	"github.com/supergiant/control/pkg/workflows/steps/ssh"
+	"github.com/supergiant/control/pkg/workflows/steps/storageclass"
 	"github.com/supergiant/control/pkg/workflows/steps/tiller"
```
stgleb: goimports

templates/storageclass.sh.tpl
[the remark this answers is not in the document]
saheienko: Local provisioner could be used as default.
https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/storage-class/local/default.yaml

@gopherstein

gopherstein:
Fixes 1147

PR #1176, opened 2018-12-07
[WIP] Add step for installing supergiant tools to a cluster

saheienko:


saheienko:
Will be redesigned later.

Issue #1044, opened 2018-11-28
2.0 UI: UX for "todo" tasks needs to be refined


gopherstein:
![Screen Shot 2018-11-28 at 7:56 PM](http://marker.screenshots.prod.s3.amazonaws.com/fd6050a9b10b4fd4b198224a8a4eb76b-1543434984213.png)

**Summary:**
Since all tasks now show up from the beginning, UX for provisioner wizard needs to be revisited

**Steps to Reproduce:** 
1. Spin up cluster
2. Open task steps or task log

**Expected Results:** 
Disallow clicking on steps or logs for todo tasks?

**Actual Results:** 
Blank screens

**Notes:**

---
### More Info

**Reported by:** Joe Kane (joe@qbox.io)
**Source URL**: [http://localhost:4200/clusters/cbe9f739](http://localhost:4200/clusters/cbe9f739)

<table><tr><td><strong>Browser</strong></td><td>Chrome 70.0.3538.110</td></tr><tr><td><strong>Screen Size</strong></td><td>1366 x 768</td></tr><tr><td><strong>OS</strong></td><td>OS X 10.14.1</td></tr><tr><td><strong>Viewport Size</strong></td><td>1680 x 971</td></tr><tr><td><strong>Zoom Level</strong></td><td>100%</td></tr><tr><td><strong>Pixel Ratio</strong></td><td>@2x</td></tr></table>

josephkane:
@hetmanchuk tagging you so you are aware of this, in case you have any ideas/opinions

hetmanchuk:
@josephkane could we hide icons until some info will appear there? or make it disabled (I could design style).

BTW I made some improvements for this section, added animated title, issue #954

PR #1178, opened 2018-12-07
Pre Provision Task

josephkane:
* handle new pre-provision task

* fix long polling error

* disable steps and logs button when task is todo


gopherstein:
Fixes #1044

PR #1189, opened 2018-12-10
Add unit tests for create instance

stgleb:


pkg/workflows/steps/digitalocean/create_instance_test.go
```diff
@@ -0,0 +1,163 @@
...
+						},
+						{
+							Type:      "private",
+							IPAddress: "5.6.7.8",
+						},
+					},
+				},
+			},
+			dropletErr:  nil,
+			expectError: false,
```
saheienko: No need to set default values.

Issue #953, opened 2018-11-20
2.1: UI - Master size selection should generate feedback


eldridgeh:
**Short Summary:**
Master size selection should generate feedback to show how many nodes the cluster could handle.
example:
If I select 1 m4.4xlarge master, it should suggest that it could handle 1000 nodes.
If I selected 2 m4.4xlarge masters, it should suggest that it could handle 2000 nodes and so on.

gopherstein:
combining these issues. https://github.com/supergiant/control/issues/843

PR #1200, opened 2018-12-11
feat (add-node.component) add sorting for machine types for DO

nikage:


eldridgeh:
@nikage Can this be improved on? Thanks

nikage:
@eldridgeh I think I'm gonna rewrite the sorting algorithm from scratch)

josephkane:
@nikage would it be possible to sort DO machine types by their size? I know we get that information back, but I wonder how hard that would be?

josephkane:
@nikage I'm not sure what you mean, but that prop is meant to be a reference for the resources available for each of the different machine types. Ideally in the future, we would use that information to 1. build pre-set configurations for different cluster uses (video processing, data store, etc.), and 2. display feedback on the review step (https://github.com/supergiant/control/issues/953). I had thought that using these sizes might be an easier way to sort DO machines, but it looks like you figured out a way to do it without them. Does that make sense, or did I misunderstand your question?

eldridgeh:
@nikage 
I need to close this PR?

nikage:
@eldridgeh approve it if you please

PR #1224, opened 2018-12-19
Cover profile package with tests

stgleb:


pkg/profile/profile.go
```diff
@@ -9,7 +9,7 @@ type Profile struct {
 	NodesProfiles  []NodeProfile `json:"nodesProfiles" valid:"-"`
 
 	// StaticAuth represents tokens and basic authentication credentials that
-	// would be set to kube-apiserver on start.
+	// would be set to profile-apiserver on start.
```
saheienko: Use `kube` prefix here.

stgleb: @saheienko done

PR #1232, opened 2018-12-23
Cover kube with unit tests

stgleb:


pkg/kube/handler.go
```diff
@@ -11,6 +11,8 @@ import (
 	"sync"
 	"time"
 
+	core "k8s.io/api/core/v1"
```
saheienko: It's more common to use add version info to aliases(`corev1`).

pkg/kube/handler.go
```diff
@@ -1139,24 +1121,25 @@ func (h *Handler) getServices(w http.ResponseWriter, r *http.Request) {
 	var targetServices = make([]*proxy.Target, 0)
 
 	for _, service := range k8sServices.Items {
-		if !contains("kubernetes.io/cluster-service", "true", service.Metadata.Labels) {
+		if !contains("kubernetes.io/cluster-service",
```
saheienko: Please, use it as a constant.

pkg/kube/handler.go
```diff
@@ -1139,24 +1121,25 @@ func (h *Handler) getServices(w http.ResponseWriter, r *http.Request) {
...
 			continue
 		}
 
 		if len(service.Spec.Ports) == 1 && service.Spec.Ports[0].Protocol == "TCP" {
 			var serviceInfo = &ServiceInfo{
-				ID:        service.Metadata.UID,
-				Name:      service.Metadata.Name,
-				Type:      service.Spec.Type,
-				Namespace: service.Metadata.Namespace,
+				ID:        string(service.ObjectMeta.UID),
```
saheienko: Is it possible to use a service struct instead of converting to a ServiceInfo one?

stgleb: Well, this is possible, but this will require change from other control components and analyze. I think this is out of scope of this particular PR.

saheienko: Yes, TODO description would be enough for this PR.

PR #1233, opened 2018-12-24
Refactor package message

stgleb:
Add unit tests
Remove dead codepaths


pkg/message/message.go
```diff
@@ -31,12 +29,7 @@ func New(userMessage string, devMessage string, code sgerrors.ErrorCode, moreInf
 	}
 }
 func SendMessage(w http.ResponseWriter, msg Message, status int) {
-	data, err := json.Marshal(msg)
-	if err != nil {
-		logrus.Errorf("failed to marshall message: %v", err)
-		http.Error(w, "", http.StatusInternalServerError)
-		return
-	}
+	data, _ := json.Marshal(msg)
```
saheienko: Please, check the error.
Other functions use similar flow, could it be moved to the message method?

stgleb: Well, in general I agree, but it still strange for me why we check error all the time even for type that cannot fail in marshalling by design unless you add any unserializable type like chan

saheienko: It's totally bad idea to ignore errors even if it looks like it can't happen.

PR #1256, opened 2019-01-15
Drain node resources on delete

### LINKED ISSUE #1181: 2.0.1: API - Deleting a node should first drain the node
**Short Summary:**
When we delete a node from the cluster, we should first drain the node to allow time for all resources to be rescheduled. We currently simply terminate the instance.

stgleb:
Drain node resources before delete

Closes #1181

### DIFF
--- a/templates/drain.sh.tpl
+++ b/templates/drain.sh.tpl
@@ -1,0 +1,1 @@
+kubectl drain {{ .NodeName }} --ignore-daemonsets --force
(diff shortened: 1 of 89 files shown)


templates/drain.sh.tpl
```diff
@@ -0,0 +1,3 @@
+sudo kubectl drain $(sudo kubectl get no -o wide|grep {{ .PrivateIP }}| awk '{ print $1 }') \
```
saheienko: Is it possible to do this with the kubernetes client?

stgleb: Well, I thought about eviction API, but found it more complicated compare to ssh-ing to a node and running kubectl. Talking about client I've not found implementation for draining node in client.
Maybe I could contribute to it, but I am not sure.

https://github.com/kubernetes-client/python/issues/188

PR #1257, opened 2019-01-15
Allow API consumers to add root user on cold start. 

egopher:
Implemented cold start determination endpoint. Enabled API consumer to register root user at the coldstart.
The cold start is defined as situation when there is 0 users in the database.


pkg/user/service_test.go
```diff
@@ -3,6 +3,7 @@ package user
 import (
 	"context"
 	"encoding/json"
```
stgleb: goimports

pkg/controlplane/server.go
```diff
@@ -277,7 +242,7 @@ func configureApplication(cfg *Config) (*mux.Router, error) {
 	if err != nil {
 		return nil, errors.Wrap(err, "new helm service")
 	}
-	go ensureHelmRepositories(helmService)
+	go ensureHelmRepositories(helmService, userService)
```
saheienko: No need to pass `userService` to the function, check for a cold start here.

pkg/user/handler.go
```diff
@@ -66,6 +66,57 @@ func (h *Handler) Authenticate(w http.ResponseWriter, r *http.Request) {
...
+	if coldstart {
+		if err := h.userService.Create(r.Context(), &user); err != nil {
+			message.SendUnknownError(w, err)
+			return
+		}
+	} else {
+		message.SendMessage(w, message.Message{
+			UserMessage: "Root user already exists",
+			ErrorCode:   sgerrors.EntityAlreadyExists,
+		}, http.StatusForbidden)
```
saheienko: Is it an auth error?
`message.SendAlreadyExists` uses `http.StatusConflict`.

pkg/controlplane/server.go
```diff
@@ -52,6 +49,7 @@ import (
 	"github.com/supergiant/control/pkg/workflows/steps/ssh"
 	"github.com/supergiant/control/pkg/workflows/steps/storageclass"
 	"github.com/supergiant/control/pkg/workflows/steps/tiller"
+	"k8s.io/helm/pkg/repo"
```
stgleb: goimports

pkg/user/handler.go
```diff
@@ -5,10 +5,9 @@ import (
 	"fmt"
 	"net/http"
 
-	"gopkg.in/asaskevich/govalidator.v8"
-
 	"github.com/supergiant/control/pkg/message"
 	"github.com/supergiant/control/pkg/sgerrors"
+	"gopkg.in/asaskevich/govalidator.v8"
```
stgleb: goimports

pkg/controlplane/server.go
egopher: It's was goimported.

pkg/user/handler.go
egopher: It's was goimported.

PR #1259, opened 2019-01-16
Move ssh config to the Kube model

saheienko:
Control uses different types for cluster bootstrapping and representation; the idea here to use kube model for both. 
This change is a part of merging a provisioner config and a cluster model.

### DIFF
--- a/pkg/provisioner/provisioner.go
+++ b/pkg/provisioner/provisioner.go
@@ -535,21 +535,17 @@ func (tp *TaskProvisioner) buildInitialCluster(ctx context.Context,
 	config *steps.Config, taskIds map[string][]string) error {
 
 	cluster := &model.Kube{
-		ID:                 config.ClusterID,
-		State:              model.StatePrepare,
-		Name:               config.ClusterName,
-		Provider:           profile.Provider,
-		AccountName:        config.CloudAccountName,
-		RBACEnabled:        profile.RBACEnabled,
-		ServicesCIDR:       profile.K8SServicesCIDR,
-		Region:             profile.Region,
-		Zone:               profile.Zone,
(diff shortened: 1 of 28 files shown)


pkg/provisioner/provisioner.go
```diff
@@ -535,21 +535,17 @@ func (tp *TaskProvisioner) buildInitialCluster(ctx context.Context,
...
+		State:        model.StateProvisioning,
+		Name:         config.ClusterName,
+		Provider:     profile.Provider,
+		AccountName:  config.CloudAccountName,
+		RBACEnabled:  profile.RBACEnabled,
+		ServicesCIDR: profile.K8SServicesCIDR,
+		Region:       profile.Region,
+		Zone:         profile.Zone,
+		User:         profile.User,
+		Password:     profile.Password,
```
stgleb: Maybe put it to Kube model too?

saheienko: If you mean User&Password these are deprecated ones.

pkg/provisioner/provisioner.go
```diff
@@ -577,6 +573,8 @@ func (tp *TaskProvisioner) buildInitialCluster(ctx context.Context,
 		Masters:   masters,
 		Nodes:     nodes,
 		Tasks:     taskIds,
+
```
stgleb: Maybe use config.Kube to save to etcd instead of creating new one once again? Just fill it with the data from profile.

saheienko: config.Kube is only used for storing ssh configuration for now.

stgleb: Okay, but for ongoing changes I would prefer to move there certificates config either. All things that belongs to kube directly.

PR #1267, opened 2019-01-21
Gce refine

josephkane:


josephkane:
* gce implementation has been in for awhile

* remove 'experimental' tag

* rework cloud account list provider names to be congruent

PR #1268, opened 2019-01-22
Introduce provider steps to provisioner workflows

saheienko:
All cluster provisioning worflows (awsMaster, gceMaster etc) use the same set of steps except of the working with the cloud infrastructure ones.

Add provider steps to unify interaction with different infrastructure providers.

### DIFF
--- a/pkg/workflows/steps/provider/cleanup.go
+++ b/pkg/workflows/steps/provider/cleanup.go
@@ -1,0 +1,83 @@
+package provider
+
+import (
+	"context"
+	"fmt"
+	"io"
+
+	"github.com/pkg/errors"
+
+	"github.com/supergiant/control/pkg/clouds"
+	"github.com/supergiant/control/pkg/workflows/steps"
+	"github.com/supergiant/control/pkg/workflows/steps/amazon"
+	"github.com/supergiant/control/pkg/workflows/steps/digitalocean"
+	"github.com/supergiant/control/pkg/workflows/steps/gce"
+)
+
+const (
+	CleanUpStep = "cleanUp"
+)
+
+type StepCleanUp struct {
+}
+
+func (s StepCleanUp) Run(ctx context.Context, out io.Writer, cfg *steps.Config) error {
+	if cfg == nil {
+		return errors.New("invalid config")
+	}
+
+	steps, err := cleanUpStepsFor(cfg.Provider)
+	if err != nil {
+		return errors.Wrap(err, CleanUpStep)
+	}
+	for _, s := range steps {
+		if err = s.Run(ctx, out, cfg); err != nil {
+			return errors.Wrap(err, CleanUpStep)
+		}
+	}
+
+	return nil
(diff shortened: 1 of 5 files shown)


pkg/workflows/steps/provider/cleanup.go
```diff
@@ -0,0 +1,85 @@
...
+			steps.GetStep(amazon.DeleteSecurityGroupsStepName),
+			steps.GetStep(amazon.DisassociateRouteTableStepName),
+			steps.GetStep(amazon.DeleteSubnetsStepName),
+			steps.GetStep(amazon.DeleteRouteTableStepName),
+			steps.GetStep(amazon.DeleteInternetGatewayStepName),
+			steps.GetStep(amazon.DeleteKeyPairStepName),
+			steps.GetStep(amazon.DeleteVPCStepName),
+		}, nil
+	case clouds.DigitalOcean:
+		return []steps.Step{
```
stgleb: why don't you use steps.GetStep here?

pkg/workflows/workflow.go
```diff
@@ -129,15 +90,14 @@ func Init() {
 		steps.GetStep(poststart.StepName),
 	}
 
-	awsNodeWorkflow := []steps.Step{
-		steps.GetStep(amazon.StepNameCreateEC2Instance),
+	nodeWorkflow := []steps.Step{
```
stgleb: Well, originally I thought about having  this step in steps package as before, but call provider.CreateMachine for actual work.

pkg/workflows/workflow.go
```diff
@@ -151,88 +111,35 @@ func Init() {
...
-	workflowMap[AWSMaster] = awsMasterWorkflow
-	workflowMap[AWSNode] = awsNodeWorkflow
-	workflowMap[AWSPreProvision] = awsPreProvision
-	workflowMap[AWSDeleteCluster] = awsDeleteClusterWorkflow
-	workflowMap[AWSDeleteNode] = awsDeleteNodeWorkflow
-	workflowMap[GCENode] = gceNodeWorkflow
-	workflowMap[GCEMaster] = gceMasterWorkflow
-	workflowMap[GCEDeleteCluster] = gceDeleteCluster
-	workflowMap[GCEDeleteNode] = gceDeleteNode
+
```
stgleb: Well, no this workflowMap is obsolete since steps for creating, deleting machines are the same. We need to refactor Provisioner and this map to be provider agnostic.

saheienko: Yes, it requires changes in provisioner. Should it be in this PR?

pkg/workflows/steps/provider/cleanup.go
saheienko: Fixed.

pkg/workflows/workflow.go
stgleb: I am okay with that.

PR #1271, opened 2019-01-25
Add file and memory storage

stgleb:


cmd/controlplane/main.go
```diff
@@ -18,7 +18,8 @@ var (
 	version       = "undefined"
 	addr          = flag.String("address", "0.0.0.0", "network interface to attach server to")
 	port          = flag.Int("port", 8080, "tcp port to listen for incoming requests")
-	etcdURL       = flag.String("etcd-url", "localhost:2379", "etcd url with port")
+	storageType   = flag.String("storageType", "file", "storage type either file(default), memory or etcd")
```
egopher: storage-mode should be better name.

cmd/controlplane/main.go
```diff
@@ -18,7 +18,8 @@ var (
 	version       = "undefined"
 	addr          = flag.String("address", "0.0.0.0", "network interface to attach server to")
 	port          = flag.Int("port", 8080, "tcp port to listen for incoming requests")
-	etcdURL       = flag.String("etcd-url", "localhost:2379", "etcd url with port")
+	storageMode   = flag.String("storageMode", "file", "storage type either file(default), memory or etcd")
```
egopher: please make params snake case instead of camel case.

stgleb: storage_mode? this doesn't fix to unix cli arg naming convention.

PR #1273, opened 2019-01-28
Refactor node package

saheienko:
* move Node struct to the model package;
* rename Node to Machine.


saheienko:
@stgleb API and models are the same, how does it break UI?

PR #1278, opened 2019-01-31
Restart failed cluster

stgleb:


pkg/workflows/steps/certificates/certificates.go
```diff
@@ -45,9 +46,12 @@ func (s *Step) Run(ctx context.Context, out io.Writer, config *steps.Config) err
...
 
 	if !config.IsMaster {
-		master := config.GetMaster()
-		config.CertificatesConfig.MasterHost = master.PrivateIp
-		config.CertificatesConfig.NodeName = config.Node.Name
+		if master := config.GetMaster(); master != nil {
+			config.CertificatesConfig.MasterHost = master.PrivateIp
+			config.CertificatesConfig.NodeName = config.Node.Name
+		} else {
+			logrus.Errorf("master not found in config")
```
saheienko: Should we continue in this case?

pkg/workflows/steps/config.go
```diff
@@ -469,15 +474,22 @@ func NewConfigFromKube(profile *profile.Profile, k *model.Kube) (*Config, error)
 		configChan:       make(chan *Config),
 	}
 
-	if k != nil {
-		cfg.Kube = *k
```
saheienko: Please, keep this check there are cases when it is `nil`.

pkg/workflows/steps/config.go
```diff
@@ -380,6 +380,9 @@ func NewConfigFromKube(profile *profile.Profile, k *model.Kube) (*Config, error)
 		DigitalOceanConfig: DOConfig{
 			Region: profile.Region,
 		},
+		Kube: *k,
```
saheienko: Could this panic with the `nil pointer dereference`?

pkg/workflows/steps/certificates/certificates.go
stgleb: I agree, lets fail here

pkg/workflows/steps/config.go
stgleb: I've added a check above

PR #1280, opened 2019-02-01
embed UI into binary

stgleb:


pkg/controlplane/server.go
```diff
@@ -145,6 +145,7 @@ func validate(cfg *Config) error {
 		return errors.New("port can't be negative")
 	}
 
+
```
saheienko: Not sure this new line is required.

PR #1284, opened 2019-02-06
Integrate kubadm to provisioner

stgleb:


pkg/workflows/steps/config.go
```diff
@@ -393,7 +315,7 @@ func NewConfig(clusterName, clusterToken, cloudAccountName string, profile profi
 }
 
 func NewConfigFromKube(profile *profile.Profile, k *model.Kube) *Config {
-	clusterToken := uuid.New()
+	token, _ := bootstrap.GenerateBootstrapToken()
```
saheienko: Please, check the error.

templates/kubelet.tpl
```diff
@@ -1,51 +1,6 @@
...
-Restart=always
-StartLimitInterval=0
-RestartSec=10
-TimeoutStartSec=30
-KillMode=process
-
-[Install]
-WantedBy=multi-user.target
+sudo bash -c "cat > /etc/default/kubelet <<EOF
+KUBELET_EXTRA_ARGS=--tls-cert-file=/etc/kubernetes/pki/kubelet.crt --tls-private-key-file=/etc/kubernetes/pki/kubelet.key
```
saheienko: It would be
```
KUBELET_EXTRA_ARGS=--tls-cert-file=/etc/kubernetes/pki/kubelet.crt
```
Use quotes to escape  whitespace symbols.

stgleb: What do you mean?

saheienko: Should it be
```
KUBELET_EXTRA_ARGS=--tls-cert-file=/etc/kubernetes/pki/kubelet.crt --tls-private-key-file=/etc/kubernetes/pki/kubelet.key
```

stgleb: Well, it works without quotes, I am afraid that quotes will not be terminated. I'll try it out

pkg/workflows/steps/config.go
stgleb: done

stgleb:
@egopher done

PR #1288, opened 2019-02-13
Add handler for listing cluster nodes

saheienko:
Add endpoint for retrieving cluster nodes info:

```
curl -XGET localhost:8080/v1/api/kubes/{kubeID}/nodes
```


pkg/kube/service.go
```diff
@@ -203,6 +211,23 @@ func (s Service) GetKubeResources(ctx context.Context, kubeID, resource, ns, nam
 	return raw, nil
 }
 
+func (s Service) ListNodes(ctx context.Context, kube *model.Kube, role string) ([]corev1.Node, error) {
```
stgleb: I think that method worth to be covered with tests

saheienko: Done.

PR #1290, opened 2019-02-15
Add CreateResourceGroup step for Azure cloud provider

saheienko:
Use [ResourceGroup](https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-group-overview#resource-groups) to deal with cluster infrastructure as with a single resource on deletion process.


stgleb:
Please update dependencies and this is good to go

PR #1298, opened 2019-02-20
Add DeleteCluster step for Azure cloud

saheienko:
Control adds all cluster resources to a particular resources group. Delete the group to remove a cluster infrastructure.

### DIFF
--- a/pkg/workflows/steps/azure/client.go
+++ b/pkg/workflows/steps/azure/client.go
@@ -1,16 +1,25 @@
 package azure
 
 import (
+	"context"
 	"fmt"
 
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -27,7 +27,7 @@ build-image:
 	docker build -t $(DOCKER_IMAGE_NAME):latest .
 
 test:
-	go test ./pkg/...
+	go test -race ./pkg/...
```
egopher: How it's related to cluster deletion?

saheienko: Removed.

PR #1307, opened 2019-03-07
[WIP] make build-ui once

stgleb:


taaraora:
I would prefer to remove lines 
```
  - make build-ui
  - statik -src=./cmd/ui/assets/dist
```
from `.travis.yml`

stgleb:
@taaraora In that case neither `run_tests.sh` nor `make build` won't work because control can't compile

taaraora:
Of course, if you remove those lines that will require to have other changes.

cmd/controlplane/main.go
```diff
@@ -36,7 +37,7 @@ func main() {
 
 	configureLogging(*logLevel, *logFormat)
 
-	cfg := &controlplane.Config{
```
taaraora: IMHO ```controlplane.Config``` is more meaningful than ```config.Config```, the first option clarifies for what exactly we create the config struct.

run_tests.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 echo "Running tests"
 
-go test -v -race -covermode=atomic -coverprofile=profile.cov ./pkg/...
+go test -v -race -covermode=atomic -tags dev -coverprofile=profile.cov ./pkg/...
```
saheienko: Do we use `dev` build tag?

travis_build.sh
```diff
@@ -35,6 +38,9 @@ elif [[ "$TRAVIS_BRANCH" == *release-* ]]; then
 	echo "unstable branch"
 	export TAG="${TAG}-unstable"
 	echo "Tag Name: ${TAG}"
+	make build-ui
+    statik -src=./cmd/ui/assets/dist
+    make build-prod
```
taaraora: You are using a nonexistent goal ```build-prod```?

travis_build.sh
```diff
@@ -35,6 +38,9 @@ elif [[ "$TRAVIS_BRANCH" == *release-* ]]; then
 	echo "unstable branch"
 	export TAG="${TAG}-unstable"
 	echo "Tag Name: ${TAG}"
+	make build-ui
+    statik -src=./cmd/ui/assets/dist
```
taaraora: I prefer to invoke ```statik``` in this else branch just once.

travis_build.sh
```diff
@@ -35,6 +38,9 @@ elif [[ "$TRAVIS_BRANCH" == *release-* ]]; then
 	echo "unstable branch"
 	export TAG="${TAG}-unstable"
 	echo "Tag Name: ${TAG}"
+	make build-ui
```
taaraora: I would prefer to build UI just once, this invocation is excessive.

travis_build.sh
```diff
@@ -18,6 +18,9 @@ echo "Tag Name: ${TAG}"
 # dockerhub, and then a release is pushed to the releases page.
 if [[ "$TRAVIS_TAG" =~ ^v[0-9]. ]]; then
 	echo "release"
+    make build-ui
```
taaraora: I would prefer to build UI just once, this invocation is excessive.

taaraora: I would prefer to build UI just once, this invocation is excessive.

run_tests.sh
stgleb: yes, we need it to build whole project. I am also not very happy with that.

taaraora:
Unfortunately ```docker_build.sh``` fails with error in this branch.

travis_build.sh
stgleb: it invokes once

travis_build.sh
stgleb: it invokes once

stgleb: it invokes once

stgleb: because only single branch of if statement is executed

taaraora: just take a look two lines below, ```docker_build.sh``` is invoked there, and where the same commands as in ```build-ui``` goal are invoked.

taaraora: see my comment below

PR #1318, opened 2019-03-15
Network provider

stgleb:
Add network support on backend

### DIFF
--- a/templates/network.sh.tpl
+++ b/templates/network.sh.tpl
@@ -478,8 +478,6 @@ sudo kubectl create -f flannel.yaml
 
 {{ if eq .NetworkProvider "Calico" }}
 sudo bash -c 'cat << EOF > rbac-kdd.yaml
-# Calico Version v3.3.2
-# https://docs.projectcalico.org/v3.3/releases#v3.3.2
 kind: ClusterRole
 apiVersion: rbac.authorization.k8s.io/v1beta1
 metadata:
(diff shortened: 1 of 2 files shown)


templates/network.sh.tpl
```diff
@@ -1102,254 +1011,5 @@ sudo kubectl create -f calico.yaml
...
-              hostPath:
-                path: /lib/modules
-            - name: xtables-lock
-              hostPath:
-                path: /run/xtables.lock
-                type: FileOrCreate
-      updateStrategy:
-        type: RollingUpdate
-EOF'
+kubectl apply -f "https://cloud.weave.works/k8s/net?k8s-version=$(kubectl version | base64 | tr -d '\n')"
```
taaraora: May be it is better to put all these entities in a separate files? But do not load from external server?
```https://cloud.weave.works``` can be  firewalled.

egopher: @stgleb Please copy it there instead of using external server.

stgleb: That doesn't work even though content is almost identical, you can try it. I've opened issue to wave github.

stgleb: I agree with you, but weave installation is weird and contain some dynamic things inside that can't be just hard coded.  I didn't dig it deeper, but when install from the saved file - it fails.

stgleb: Done

Issue #1129, opened 2018-12-04
UI becomes unresponsive


stgleb:
SG UI App works like single threaded app, when request to backend stucks for a while - no other actions on UI can be done.

Steps to reproduce

1. Deploy Kube on GCE
2. Wait a few minutes after it is deployed
3. SG logs should look like

```
time="2018-12-04T22:24:21+02:00" level=error msg="helm: list releases: 871667f7 cluster: list releases: get tiller pod: Get https://35.198.142.78/api/v1/pods?labelSelector=app%3Dhelm%2Cname%3Dtiller: dial tcp 35.198.142.78:443: i/o timeout"
```

4.Pres delete button

Expected result - cluster is deleted

Actual result - cluster gets deleted, but with substantial lag.

PR #1322, opened 2019-03-15
Remove Timeout From Cluster Apps List

josephkane:
* remove timeout from `/GET releases`

* modify communication in apps table

* SHIM: fix new cluster review step re: flannel version -> network provider

NOTE: this essentially reopens https://github.com/supergiant/control/issues/1129
However, this was a band-aid fix that severely negatively impacted UX
We should revisit the issue and fix the actual problem

### DIFF
--- a/cmd/ui/assets/src/app/clusters/new-cluster/new-cluster.component.html
+++ b/cmd/ui/assets/src/app/clusters/new-cluster/new-cluster.component.html
@@ -83,7 +83,7 @@
               <mat-option *ngFor="let n of clusterOptions.networkProviders"
                           [value]="n">{{ n }}</mat-option>
             </mat-select>
-            <mat-hint align="end">Read more about 
+            <mat-hint align="end">Read more about
               <a href="https://kubernetes.io/docs/concepts/cluster-administration/networking/"
                  target="_blank">cluster networking</a>
             </mat-hint>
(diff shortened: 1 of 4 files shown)


cmd/ui/assets/src/app/clusters/new-cluster/new-cluster.component.html
```diff
@@ -621,7 +621,7 @@ <h3>CLOUD ACCOUNT</h3>
             <h3>CLUSTER DETAILS</h3>
             <div>
               <p><span class="label">K8s VERSION:</span> {{ clusterConfig.value.K8sVersion }}</p>
-              <p><span class="label">FLANNEL VERSION:</span> {{ clusterConfig.value.flannelVersion }}</p>
+              <p><span class="label">NETWORK PROVIDER:</span> {{ clusterConfig.value.networkProvider }}</p>
```
josephkane: @kellystrauch0 we missed this spot earlier

kellystrauch0: Dangit -- good catch

taaraora:
It's great that code duplication was removed. 
If timeout still is needed it is better to add it inside ```UtilService.fetch()``` and make configurable and used at the time of dependency initialization, I mean at the time when service ```UtilService``` is being initialized.

PR #1324, opened 2019-03-18
Ignore CPU preflight checks for small machines

stgleb:


templates/kubeadm.sh.tpl
```diff
@@ -16,14 +16,14 @@ sudo systemctl restart kubelet
 sudo kubeadm config images pull
 
 {{ if .IsBootstrap }}
-sudo kubeadm init --apiserver-advertise-address={{ .AdvertiseAddress }} --token={{ .Token }} --pod-network-cidr={{ .CIDR }} \
+sudo kubeadm init --apiserver-advertise-address={{ .AdvertiseAddress }} --ignore-preflight-errors=NumCPU --token={{ .Token }} --pod-network-cidr={{ .CIDR }} \
```
saheienko: Is it possible to keep this checks for masters?

stgleb: Well, yes. But masters work fine with 1 vCPU and 1 GB of memory with reasonable 25% CPU and 75% of memory usage.

PR #1326, opened 2019-03-18
Add CreateVNet step for Azure provider

saheienko:


pkg/workflows/steps/azure/create_vnet.go
```diff
@@ -0,0 +1,79 @@
...
+
+func (s *CreateVirtualNetworkStep) Rollback(context.Context, io.Writer, *steps.Config) error {
+	return nil
+}
+
+func (s *CreateVirtualNetworkStep) Name() string {
+	return CreateVirtualNetworkStepName
+}
+
+func (s *CreateVirtualNetworkStep) Depends() []string {
```
stgleb: I wonder if it is really indpendent?

saheienko: Updated

PR #1332, opened 2019-03-22
GCE load balancers

stgleb:
Add create Target pool step


pkg/workflows/steps/config.go
```diff
@@ -60,21 +60,46 @@ type DOConfig struct {
...
+	Type      string `json:"type"`
+	ProjectID string `json:"project_id"`
+
+	PrivateKeyID string `json:"private_key_id"`
+	PrivateKey   string `json:"private_key"`
+
 	ClientEmail string `json:"client_email"`
-	TokenURI    string `json:"token_uri"`
-	ProjectID   string `json:"project_id"`
+	ClientID    string `json:"client_id"`
```
saheienko: `GCEConfig` uses camelCase for json tags. Please, use the same naming approach here.

pkg/workflows/steps/gce/common.go
```diff
@@ -18,35 +20,64 @@ type computeService struct {
...
+	deleteIpAddress            func(context.Context, steps.GCEConfig, string) (*compute.Operation, error)
 }
 
 func Init() {
-	createInstance, _ := NewCreateInstanceStep(time.Second*10, time.Minute*1)
+	createInstance, _ := NewCreateInstanceStep(time.Second*10, time.Minute*5)
 	deleteCluster, _ := NewDeleteClusterStep()
 	deleteNode, _ := NewDeleteNodeStep()
+	createTargetPool, _ := NewCreateTargetPoolStep()
+	createIPAddress, _ := NewCreateAddressStep()
```
saheienko: Please, check errors here.

pkg/workflows/steps/config.go
stgleb: I wish I can do so, bu GCE service account file uses snake case for that, I've just preserved their original formatting to deserialize json to that struct easily.

pkg/workflows/steps/gce/common.go
stgleb: I've revisited this code, errors are not needed there.

saheienko: That can be ok now but if it fail in future?

stgleb: Well, creation of Step as it is now can't fail since it doesn't do any calls to code that can return an error, but in future if steps are modified to be failure prone -  I can rewrite it.

saheienko:
Please, fix failing CI before merge.

stgleb:
@saheienko yeah, but I can't figure out what is the root cause

PR #1335, opened 2019-03-26
Security groups configuration

saheienko:
Add ability to configure ingress rules for master/node security groups on AWS.

### DIFF
--- a/pkg/workflows/steps/amazon/create_security_groups.go
+++ b/pkg/workflows/steps/amazon/create_security_groups.go
@@ -11,6 +11,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 
+	"github.com/supergiant/control/pkg/profile"
 	"github.com/supergiant/control/pkg/util"
 	"github.com/supergiant/control/pkg/workflows/steps"
 )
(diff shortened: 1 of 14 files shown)


pkg/workflows/steps/amazon/create_security_groups.go
```diff
@@ -227,6 +235,38 @@ func (s *CreateSecurityGroupsStep) whiteListSupergiantIP(ctx context.Context, EC
...
 
+func whiteListAddresses(ctx context.Context, EC2 secGroupService, groups []string, addrs []profile.Addresses) error {
+	if len(addrs) == 0 || len(addrs) == 0 {
+		return nil
+	}
+
+	perms := make([]*ec2.IpPermission, 0)
+	for _, addr := range addrs {
+		perms = append(perms, &ec2.IpPermission{
+			FromPort: aws.Int64(0),
```
taaraora: May be it would be better to whitelist just that port range which we need to expose?
like ```FromPort: aws.Int64(int64(addr.Port))``` ?

saheienko: Yes, it's better, thanks.

pkg/workflows/steps/amazon/create_security_groups.go
```diff
@@ -196,33 +196,33 @@ func (s *CreateSecurityGroupsStep) allowAllTraffic(ctx context.Context, EC2 secG
...
-		ToPort:     aws.Int64(443),
-		CidrIp:     aws.String(fmt.Sprintf("%s/32", supergiantIP)),
+		GroupId: aws.String(groupID),
+		IpPermissions: []*ec2.IpPermission{
+			{
+				FromPort: aws.Int64(8080),
+				ToPort:   aws.Int64(8080),
+				IpRanges: ips,
+			},
+			{
```
stgleb: Maybe add port 22?

pkg/workflows/steps/amazon/create_security_groups.go
```diff
@@ -196,33 +196,33 @@ func (s *CreateSecurityGroupsStep) allowAllTraffic(ctx context.Context, EC2 secG
...
 	}
 
 	_, err = EC2.AuthorizeSecurityGroupIngressWithContext(ctx, &ec2.AuthorizeSecurityGroupIngressInput{
-		GroupId:    aws.String(groupID),
-		FromPort:   aws.Int64(443),
-		ToPort:     aws.Int64(443),
-		CidrIp:     aws.String(fmt.Sprintf("%s/32", supergiantIP)),
+		GroupId: aws.String(groupID),
+		IpPermissions: []*ec2.IpPermission{
+			{
```
stgleb: I think 8080 is obsolete, we don't expose it

pkg/workflows/steps/amazon/create_security_groups.go
saheienko: It would be better to make this explicitly with the separate PR. (at the [moment](https://github.com/supergiant/control/blob/3841491010c79db269a673fd81b5d10e06891647/pkg/workflows/steps/amazon/create_security_groups.go#L132) it's exposed to all addresses)

pkg/workflows/steps/amazon/create_security_groups.go
saheienko: Done.

PR #1342, opened 2019-03-29
Hadrcode AWS instance types

saheienko:
Currently control shows reserved instances for the provided region (on cluster creation page).
Add a script to generate a list of the linux instances available on demand for each region and show this instances instead.


pkg/account/awsvms/main.go
```diff
@@ -0,0 +1,260 @@
...
+			},
+			{
+				Field: aws.String("tenancy"),
+				Type:  aws.String("TERM_MATCH"),
+				Value: aws.String("Shared"),
+			},
+			{
+				Field: aws.String("capacitystatus"),
+				Type:  aws.String("TERM_MATCH"),
+				Value: aws.String("UnusedCapacityReservation"), // https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-capacity-reservations.html
```
taaraora: What is user has fully utilized Capacity Reservation but we show him that instance type?

taaraora: the question was discussed at standup, this approach works for us.

saheienko: There would be the [Insufficient Instance Capacity](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/troubleshooting-launch.html#troubleshooting-launch-capacity) error on the CreateInstance step.

PR #1345, opened 2019-04-01
Increase disk size for gce vm

stgleb:


pkg/workflows/steps/gce/create_instance.go
```diff
@@ -151,6 +151,7 @@ func (s *CreateInstanceStep) Run(ctx context.Context, output io.Writer,
 				InitializeParams: &compute.AttachedDiskInitializeParams{
 					DiskName:    name + "-root-pd",
 					SourceImage: image.SelfLink,
+					DiskSizeGb:  30,
```
saheienko: Could it be set as a parameter?

stgleb: Well, we agreed with Mike that we set it with reasonable default, but in future we can add param for that. The problem is that it will requires updates on the UI, so we just skip it now.

PR #1346, opened 2019-04-01
Update AWS provider

saheienko:
Add ability to set a volume device name for aws instance while creating a new one.

### DIFF
--- a/pkg/clouds/aws/instances.go
+++ b/pkg/clouds/aws/instances.go
@@ -100,7 +100,7 @@ func (c *Client) CreateInstance(ctx context.Context, cfg InstanceConfig) (*ec2.I
 		},
 		BlockDeviceMappings: []*ec2.BlockDeviceMapping{
 			{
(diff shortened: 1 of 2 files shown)


pkg/clouds/aws/instances.go
```diff
@@ -100,7 +100,7 @@ func (c *Client) CreateInstance(ctx context.Context, cfg InstanceConfig) (*ec2.I
 		},
 		BlockDeviceMappings: []*ec2.BlockDeviceMapping{
 			{
-				DeviceName: aws.String("/dev/xvda"),
+				DeviceName: aws.String(cfg.VolumeDeviceName),
```
stgleb: I've just updated it in #1348 I've mounted this ebs as /dev/sda  to be root device. 
CreateInstance function, do we even us it anywhere?

saheienko: You've changed `pkg/workflows/steps/amazon/create_machine.go`.
This packages is used by capacity.

PR #1347, opened 2019-04-01
Save and expose bootstrap keys

stgleb:


pkg/model/kube.go
```diff
@@ -59,9 +59,8 @@ type Kube struct {
 	SshUser string `json:"sshUser"`
 	// DEPRECATED
 	SshPublicKey []byte `json:"sshKey"`
-	// DEPRECATED
+
 	BootstrapPublicKey []byte `json:"bootstrapPublicKey"`
-	// DEPRECATED
 	BootstrapPrivateKey []byte `json:"bootstrapPrivateKey"`
```
saheienko: These fields are deprecated (https://github.com/supergiant/control/pull/1259).
Use `SSHConfig` instead.

stgleb: so lets remove them

PR #1352, opened 2019-04-02
Deploy prometheus-operator to the kube-system namespace

saheienko:
Fixes https://github.com/supergiant/control/issues/1006

### DIFF
--- a/templates/prometheus.sh.tpl
+++ b/templates/prometheus.sh.tpl
@@ -2,6 +2,8 @@
 # it explicitly
 sudo /usr/bin/helm install stable/prometheus-operator \
     --name=prometheus-operator \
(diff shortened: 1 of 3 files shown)


templates/prometheus.sh.tpl
```diff
@@ -2,8 +2,10 @@
 # it explicitly
 sudo /usr/bin/helm install stable/prometheus-operator \
     --name=prometheus-operator \
+    --namespace=kube-system \
```
stgleb: = is need there? If not I am okay.

saheienko: helm cli has [this](https://github.com/spf13/pflag#command-line-flag-syntax) flag syntax.

PR #1355, opened 2019-04-03
Update kubeadm step

saheienko:
Properly set apiserver addresses for Azure/GCE providers.

### DIFF
--- a/pkg/workflows/steps/kubeadm/kubeadm.go
+++ b/pkg/workflows/steps/kubeadm/kubeadm.go
@@ -9,6 +9,7 @@ import (
 	"github.com/pkg/errors"
 
 	"github.com/supergiant/control/pkg/clouds"
+	"github.com/supergiant/control/pkg/sgerrors"
 	tm "github.com/supergiant/control/pkg/templatemanager"
 	"github.com/supergiant/control/pkg/workflows/steps"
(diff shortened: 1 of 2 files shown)


pkg/workflows/steps/kubeadm/kubeadm.go
```diff
@@ -47,26 +48,35 @@ func (t *Step) Run(ctx context.Context, out io.Writer, config *steps.Config) err
 	if config.KubeadmConfig.IsBootstrap {
 		config.KubeadmConfig.AdvertiseAddress = config.Node.PrivateIp
 
-		// TODO(stgleb): Remove that when all providers support Load Balancers
-		if config.Provider == clouds.AWS || config.Provider == clouds.DigitalOcean {
-			config.KubeadmConfig.ExternalDNSName = config.Node.PublicIp
-			config.KubeadmConfig.InternalDNSName = config.Node.PublicIp
```
stgleb: Well, we need do execute this code only when provider does not support load balancers, currently only AWS and DO supports that.

saheienko: These overwrites later with dns names for AWS and DO.

stgleb: As far as I know load balancers dns names are set on preprovision phase that happens before kubeadm step in node or master?

PR #1356, opened 2019-04-04
IP Whitelisting UI

kellystrauch0:
* added IP Whitelist to Cluster config step in provisioner wizard
* closes Jira card S20-825


cmd/ui/assets/src/app/clusters/new-cluster/new-cluster.component.ts
```diff
@@ -189,6 +204,8 @@ export class NewClusterComponent implements OnInit, OnDestroy {
           break;
       }
 
+console.log(newClusterData.profile);
```
josephkane: can be removed

cmd/ui/assets/src/app/clusters/new-cluster/new-cluster.component.ts
```diff
@@ -529,19 +546,55 @@ export class NewClusterComponent implements OnInit, OnDestroy {
...
     if (this.regionsFilter === '') {
       return val.name;
     }
 
     return val.name.toLowerCase().indexOf(this.regionsFilter.toLowerCase()) > -1;
-  };
+  }
+
+  multiLineStringToArray (str) {
+    //handle empty string
```
josephkane: comments other than TODOs can be removed

cmd/ui/assets/src/app/clusters/new-cluster/new-cluster.component.ts
```diff
@@ -529,19 +546,55 @@ export class NewClusterComponent implements OnInit, OnDestroy {
...
   regionsFilterCallback = (val) => {
     if (this.regionsFilter === '') {
       return val.name;
     }
 
     return val.name.toLowerCase().indexOf(this.regionsFilter.toLowerCase()) > -1;
-  };
+  }
+
+  multiLineStringToArray (str) {
```
taaraora: Regarding this signature. TypeScript's compiler can check types of passed arguments, etc...
Please consider adding types for arguments, returning values, variables, etc....
In general, It is better to add access modifier explicitly.
For example ```private toArray(multiLineText : string) : string[]``` or ```private toArray(multiLineText : string) : Array<string>```

cmd/ui/assets/src/app/clusters/new-cluster/new-cluster.component.ts
```diff
@@ -529,19 +546,55 @@ export class NewClusterComponent implements OnInit, OnDestroy {
...
+    //handle empty string
+    if(str.length==0){
```
taaraora: I think this method's body can be written a little bit simpler. 
Also, I think we don't need blank lines, it is better to filter them. E. g. 
```typescript
private splitLines(multiLineText : string) : Array<string> {
    return multiLineText.split("\n").filter(line => line.trim() != '');
  }
```

cmd/ui/assets/src/app/clusters/new-cluster/new-cluster.component.ts
```diff
@@ -170,6 +176,15 @@ export class NewClusterComponent implements OnInit, OnDestroy {
       newClusterData.profile.masterProfiles = this.nodesService.compileProfiles(this.selectedCloudAccount.provider, this.machines, 'Master');
       newClusterData.profile.nodesProfiles = this.nodesService.compileProfiles(this.selectedCloudAccount.provider, this.machines, 'Node');
 
+      //format exposedArray as an array of objects
+      let exposedAddrArray = this.multiLineStringToArray(newClusterData.profile.exposedAddresses);
```
taaraora: May be it can be written a little bit simplier?
```typescript
let exposedAddrObjects :Array<CIDR> = newClusterData.profile.exposedAddresses
  .split('\n')
  .filter(line => line.trim() != '')
  .map((line: string) => {cidr : line});
```
ofc this example requires new type for CIDR model to be declared 
```typescript
export interface CIDR{
  cidr : string
}
```
But that is power of TypeScript, we have type checks at compile time.

cmd/ui/assets/src/app/clusters/new-cluster/new-cluster.component.ts
```diff
@@ -529,19 +546,55 @@ export class NewClusterComponent implements OnInit, OnDestroy {
     };
   }
 
+  allValidCidrs(): ValidatorFn {
+    return (userCidrs: AbstractControl): { [key: string]: any } | null => {
```
taaraora: Maybe it is better to use method signature exactly as it was defined in the interface?
```typescript
export interface ValidatorFn {
    (control: AbstractControl): ValidationErrors | null;
}
```
consequently we can change a little bit signature of this method
```typescript
  private allValidCidrs(): ValidatorFn {
    return (userCidrs: AbstractControl): ValidationErrors | null => {

      // is true only if each cidr has being validated successfully
     let cidrValidationSuccessful = this.toArray(userCidrs.value)
        .every((cidr: string) : boolean => cidrRegex({ exact: true }).test(cidr) );

      return cidrValidationSuccessful ? null : { 'invalidCidrs': { value: true } };
    };
  }
```

kellystrauch0:
Sorry, just saw the changed requested by @taaraora .  Looking at those now.

cmd/ui/assets/src/app/clusters/new-cluster/new-cluster.component.ts
```diff
@@ -73,6 +73,8 @@ export class NewClusterComponent implements OnInit, OnDestroy {
   machineTypesFilter = '';
   regionsFilter = '';
 
+  exposedAddressesArray: Array<any>;
```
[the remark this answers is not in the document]
kellystrauch0: Good suggestion -- added

cmd/ui/assets/src/app/clusters/new-cluster/new-cluster.component.ts
kellystrauch0: Good suggestion -- added

cmd/ui/assets/src/app/clusters/new-cluster/new-cluster.component.ts
kellystrauch0: The filter() is already taking out blank lines, so I think adding trim() would be redundant.  You're right that all my if/else's were unnecessary though -- I've removed them.

PR #1357, opened 2019-04-05
Add a load balancer for masters on Azure

saheienko:


stgleb:
Please merge with current master and we are good to go

PR #1360, opened 2019-04-05
Fix add node kubeadm

stgleb:
Save bootstrap token in kube
Restore that token when adding node

### DIFF
--- a/pkg/kube/handler.go
+++ b/pkg/kube/handler.go
@@ -531,8 +531,8 @@ func (h *Handler) addMachine(w http.ResponseWriter, r *http.Request) {
 	kubeID := vars["kubeID"]
 	k, err := h.svc.Get(r.Context(), kubeID)
 
-	if err != nil {
-		http.Error(w, err.Error(), http.StatusNotFound)
+	if sgerrors.IsNotFound(err) {
+		http.NotFound(w, r)
 		return
 	}
 
--- a/pkg/kube/handler.go
+++ b/pkg/kube/handler.go
@@ -531,15 +531,28 @@ func (h *Handler) addMachine(w http.ResponseWriter, r *http.Request) {
 	kubeID := vars["kubeID"]
 	k, err := h.svc.Get(r.Context(), kubeID)
 
-	// TODO(stgleb): This method contains a lot of specific stuff, implement provision node
-	// method for nodeProvisioner to do all things related to provisioning and saving cluster state
-	if sgerrors.IsNotFound(err) {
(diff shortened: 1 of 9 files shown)


pkg/workflows/steps/amazon/create_machine.go
```diff
@@ -144,16 +144,15 @@ func (s *StepCreateInstance) Run(ctx context.Context, w io.Writer, cfg *steps.Co
...
-		runInstanceInput.NetworkInterfaces = []*ec2.InstanceNetworkInterfaceSpecification{
-			{
-				DeviceIndex:              aws.Int64(0),
-				AssociatePublicIpAddress: aws.Bool(true),
-				DeleteOnTermination:      aws.Bool(true),
-				SubnetId:                 aws.String(cfg.AWSConfig.Subnets[cfg.AWSConfig.AvailabilityZone]),
-				Groups:                   []*string{secGroupID},
-			},
-		}
+
```
stgleb: every node has public IP address by default, so skip this if statement

templates/kubeadm.sh.tpl
```diff
@@ -16,12 +16,12 @@ sudo systemctl restart kubelet
 sudo kubeadm config images pull
 
 {{ if .IsBootstrap }}
-sudo kubeadm init --ignore-preflight-errors=NumCPU --apiserver-advertise-address={{ .AdvertiseAddress }} --token={{ .Token }} --pod-network-cidr={{ .CIDR }} \
```
stgleb: By default token is valid for 24 hours, but we need token that allows to join nodes for longer period.

pkg/kube/handler.go
```diff
@@ -531,15 +531,28 @@ func (h *Handler) addMachine(w http.ResponseWriter, r *http.Request) {
...
 		return
 	}
 
+	logrus.Debugf("Get cloud profile %s", k.ProfileID)
+	kubeProfile, err := h.profileSvc.Get(r.Context(), k.ProfileID)
+
 	if err != nil {
-		http.Error(w, err.Error(), http.StatusInternalServerError)
+		if sgerrors.IsNotFound(err) {
+			message.SendNotFound(w, k.ProfileID, err)
```
saheienko: This is kubes handler. Should it be status `Bad Request` if it doesn't found profile id?

stgleb: but 404 give better clue on what happens. 400 bad request is mostly associated with bad formatted data.

PR #1364, opened 2019-04-08
Create Azure Cluster UI

josephkane:
* add Azure support to UI

* SHIM: bandaid fix for digital ocean machine size sorting

### DIFF
--- a/cmd/ui/assets/src/app/clusters/new-cluster/new-cluster.component.html
+++ b/cmd/ui/assets/src/app/clusters/new-cluster/new-cluster.component.html
@@ -61,7 +61,7 @@
     <mat-step label="Cluster Configuration"
               [stepControl]="clusterConfig">
       <form [formGroup]="clusterConfig"
-            class="cluster-config-form">
+            class="cluster-and-provider-config">
         <div class="input-group">
           <mat-form-field>
             <mat-select placeholder="K8s Version"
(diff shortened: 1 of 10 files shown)


cmd/ui/assets/src/app/clusters/new-cluster/new-cluster.component.html
```diff
@@ -322,6 +322,57 @@ <h1>Create Kubernetes Cluster</h1>
...
+          <div class="input-group public-key-wrapper">
+            <mat-form-field>
+            <textarea cdkTextareaAutosize
+                      cdkAutosizeMaxRows="11"
+                      matInput
+                      formControlName="publicKey"
+                      placeholder="Public ssh key"
+                      required></textarea>
+              <mat-hint align="end">
+                <a href="https://en.wikipedia.org/wiki/Secure_Shell#Key_management"
```
taaraora: I think it is better to add section on our docs website - https://supergiant.readme.io/docs/control-installation
Where we can tell user how exactly the one can generate keys on different OSs.

josephkane: @taaraora I definitely agree. The original intent here was to have a quick tip on what all these different fields are (including the cluster config and provider config), with a link to our docs that explains more in-depth what the field is used for, what the different options are, etc. I guess this was an attempt to do that pre-Seattle and pre-training contracts. If we have time, I think it would be great if you and I could spearhead that effort (_maybe_ even for Barcelona?) and do it well. We could maybe implement with the provisioner wizard redesign?

cmd/ui/assets/src/app/clusters/new-cluster/new-cluster.component.ts
```diff
@@ -135,6 +135,13 @@ export class NewClusterComponent implements OnInit, OnDestroy {
     }
   }
 
+  get azureVNetCIDR() {
+    if (this.selectedCloudAccount && this.selectedCloudAccount.provider == 'azure') {
+      return this.providerConfig.get('azureVNetCIDR');
+    } else {
+      return true;
```
taaraora: ```azureVNetCIDR``` is boolean?

Issue #78, opened 2016-05-06
Graceful ETCD check.


gopherstein:
During api boot. If etcd (or any other required resource) is not available. the api needs to go into a check and wait for x second mode. with a 15 min timeout. This will ensure that if there is a issue, the pod does not quick before the user can check logs for the issue. Currently the api will immediately panic if etcd is not found.

PR #1383, opened 2019-04-18
Allow editing of GCE cloud accounts

kellystrauch0:
* adds code editor to edit screen for GCE cloud accounts so that GCE creds can be changed
* resolves Jira card S20-926

## Type

- [x] Feature (**new functionality** was added)

## Details
allows editing of GCE cloud accounts

Example: Introduces a new cloud provider for provisioning and managing kubes on GCE, due to popular demand.

Example: Fixes #78, #89, #123

### Additional Details?

Example: In order to take advantage of this new feature, users will need to restart SG Control with the `gce` provider option added to the `providers` flag: `--providers=gce,aws,etc.` Otherwise, functionality is not changed.

## Checklist

### DIFF
--- a/cmd/ui/assets/src/app/system/cloud-accounts/edit-cloud-account/edit-cloud-account.component.ts
+++ b/cmd/ui/assets/src/app/system/cloud-accounts/edit-cloud-account/edit-cloud-account.component.ts
@@ -44,7 +44,17 @@ export class EditCloudAccountComponent implements OnInit, OnDestroy {
       this.saving = true;
 
       if(this.account.provider == "gce"){
-        this.account.credentials = JSON.parse(this.gceCreds);
+        try{
+          this.account.credentials = JSON.parse(this.gceCreds);
+        } catch(e) {
+          this.notifications.display(
+            'error',
+            'Account: ' + this.accountName,
+            'Error: could not parse JSON'
+          );
+          this.saving = false;
+          return;
+        }
       }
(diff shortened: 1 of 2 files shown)


cmd/ui/assets/src/app/system/cloud-accounts/edit-cloud-account/edit-cloud-account.component.ts
```diff
@@ -37,6 +43,20 @@ export class EditCloudAccountComponent implements OnInit, OnDestroy {
     if (!this.saving) {
       this.saving = true;
 
+      if(this.account.provider == "gce"){
+        try{
+          this.account.credentials = JSON.parse(this.gceCreds);
+        } catch(e) {
+          this.notifications.display(
```
kellystrauch0: @josephkane I'm displaying my own, custom error message here rather than using the displayError() function because displayError() is expecting an Angular error as its second argument (I think), and I'm only able to give it a JavaScript error (e).  That said, this seems hacky, so let me know if there's a better way.

josephkane: this is exactly right. we could probably stand to remove the `displayError()` and `displaySuccess()` seeing as with this PR we're not just using them for HTTP errors, so maybe we can do that when we implement a new notification system.

PR #1396, opened 2019-04-26
Use upload certs feature of kubeadm

stgleb:
Upload certificates to configmap to distribute them securely

- [x] Feature (**new functionality** was added)
- [x] Breaking (**existing functionality** was affected)

This change will break suppot of versions prior to 1.14
Needs changes on the UI

- [ ] I understand [the Contribution Guidelines](https://supergiant.readme.io/docs/guidelines).


templates/kubeadm.go
```diff
@@ -68,7 +68,8 @@ EOF"
 
 sudo kubeadm init --ignore-preflight-errors=NumCPU \
 --node-name ${HOSTNAME} \
---config=/etc/supergiant/kubeadm.conf
+--config=/etc/supergiant/kubeadm.conf \
+--experimental-upload-certs
```
saheienko: Should it be `--upload-certs`?

pkg/workflows/steps/config.go
```diff
@@ -384,6 +404,8 @@ func NewConfig(clusterName, cloudAccountName string, profile profile.Profile) (*
 			RBACEnabled: profile.RBACEnabled,
 		},
 		KubeadmConfig: KubeadmConfig{
+			// TODO(stgleb): get it from available versions once we have them
+			KubeadmVersion:     "1.14.3",
```
saheienko: Could we use `1.15.0`?

stgleb: @saheienko we will, but this will require changes in kubeadm step, remove experimental prefix for keys related to HA setup. I'll upgrade kubeadm  version and step template to later pull requests.

templates/kubeadm.go
stgleb: We will introcude 1.15 support when first patch arrives, so far this is unrelated changes.

PR #1426, opened 2019-05-13
Update aws provider

saheienko:
Fix configuration for instance network interfaces.
This package is used by capacity.


pkg/clouds/aws/instances.go
```diff
@@ -108,28 +108,24 @@ func (c *Client) CreateInstance(ctx context.Context, cfg InstanceConfig) (*ec2.I
 				},
 			},
 		},
```
stgleb: Why you delete that?

saheienko: It conflicts with network interface one.

PR #1433, opened 2019-05-14
Update configmap step

saheienko:
## Type

- [x] Fix (a **bug** or other **issue** was fixed)
- [ ] Feature (**new functionality** was added)

## Details

Create a `kube-system/capacity` configmap with capacity configuration. This provides an easy capacity integration.

## Checklist

- [ ] [Documentation](https://supergiant.readme.io/v2.1.0-alpha/) was **not** needed.

### DIFF
--- a/pkg/provisioner/provisioner.go
+++ b/pkg/provisioner/provisioner.go
@@ -10,21 +10,22 @@ import (
 	"sync"
 	"time"
 
-	"k8s.io/client-go/util/keyutil"
-
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
+	"k8s.io/client-go/util/keyutil"
 
-	"github.com/supergiant/control/pkg/runner/dry"
-	"github.com/supergiant/control/pkg/storage/memory"
+	"github.com/supergiant/control/pkg/clouds"
 	"github.com/supergiant/control/pkg/model"
 	"github.com/supergiant/control/pkg/pki"
 	"github.com/supergiant/control/pkg/profile"
+	"github.com/supergiant/control/pkg/runner/dry"
 	"github.com/supergiant/control/pkg/sgerrors"
 	"github.com/supergiant/control/pkg/storage"
+	"github.com/supergiant/control/pkg/storage/memory"
 	"github.com/supergiant/control/pkg/util"
 	"github.com/supergiant/control/pkg/workflows"
 	"github.com/supergiant/control/pkg/workflows/steps"
+	"github.com/supergiant/control/pkg/workflows/steps/configmap"
 )
 
 const (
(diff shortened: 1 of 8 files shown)


pkg/workflows/steps/configmap/configmap.go
```diff
@@ -1,27 +1,56 @@
...
 	clientcorev1 "k8s.io/client-go/kubernetes/typed/core/v1"
 
 	"github.com/supergiant/control/pkg/kubeconfig"
-	"github.com/supergiant/control/pkg/model"
 	"github.com/supergiant/control/pkg/workflows/steps"
 )
 
+const (
+	// userdataPrefix is used to set script interpreter and log the output.
+	// https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/user-data.html#user-data-shell-scripts
```
stgleb: I would rather name it header, but doesn't matter

pkg/provisioner/provisioner.go
```diff
@@ -594,9 +595,16 @@ func buildNodeProvisionScript(ctx context.Context, config *steps.Config) {
 	// These values must still be templated
 	dryConfig.Node.PublicIp = "{{ .PublicIp }}"
 	dryConfig.Node.PrivateIp = "{{ .PrivateIp }}"
+	dryConfig.ExternalDNSName = fmt.Sprintf("{{ .%s }}", configmap.KubeAddr)
 	dryConfig.Runner = dryRunner
 	dryConfig.DryRun = true
```
stgleb: Could add some comments on that, I see that we can determine private and public ip of the instance, but I can't get why we use it over ip's that goes from instance object

pkg/workflows/steps/configmap/configmap.go
```diff
@@ -126,30 +143,36 @@ func (s *Step) Depends() []string {
...
+	out := bytes.TrimSpace(w.Bytes())
+	if !bytes.HasPrefix(out, []byte("#!")) {
+		// https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/user-data.html#user-data-shell-scripts
+		out = append([]byte(userdataPrefix), out...)
 	}
+	return out, nil
+}
 
-	return k8sClient, nil
+func getSubnet(azSubnets map[string]string) string {
```
stgleb: If we pick up random subnet - we can also sen AZ. subnet map contains az as a key

pkg/provisioner/provisioner.go
saheienko: Userdata is a part of run-instance configuration it should be applied before instance creation.

pkg/workflows/steps/configmap/configmap.go
saheienko: Capacity has no AZ option.

pkg/provisioner/provisioner.go
stgleb: Oh, thanks

pkg/workflows/steps/configmap/configmap.go
stgleb: I've meant that subnet is local to specific az. So you can return AZ too and store it in the  config for future use, I am not sure if it will be ever needed, but why not.

PR #1453, opened 2019-05-16
Redesign no AWS account cluster import view

kellystrauch0:
* implements redesign for what user sees at Cluster Import if they have no AWS accounts
* resolves card S20-1052

## Type

- [ ] Feature (**new functionality** was added)
- [x] Other (please **explain** below)
UI only

## Details

Example: Introduces a new cloud provider for provisioning and managing kubes on GCE, due to popular demand.

Example: Fixes #78, #89, #123

### Additional Details?

Example: In order to take advantage of this new feature, users will need to restart SG Control with the `gce` provider option added to the `providers` flag: `--providers=gce,aws,etc.` Otherwise, functionality is not changed.

## Checklist


cmd/ui/assets/src/app/clusters/import-cluster/import-cluster.component.scss
```diff
@@ -2,33 +2,45 @@
 
   .warning-wrapper {
     display: flex;
+    flex-direction: row;
+    justify-content: center;
+    margin-top: 38px;
```
kellystrauch0: @josephkane This is a hard-coded margin from the top -- would it be better to center everything between the header and footer instead?

josephkane: I think this might be to avoid collisions with the top nav positioning. If thats not the case, then yes it would definitely be better.

kellystrauch0: After in-person discussion w/ @josephkane, this is okay as is.

PR #1455, opened 2019-05-17
Do not fail if drain node fails

stgleb:
Skip error on drain step failure

### DIFF
--- a/pkg/workflows/steps/drain/drain_node.go
+++ b/pkg/workflows/steps/drain/drain_node.go
@@ -3,12 +3,12 @@ package drain
 import (
 	"context"
 	"fmt"
-	"github.com/sirupsen/logrus"
 	"io"
 	"text/template"
 
 	"github.com/pkg/errors"
 
+	"github.com/sirupsen/logrus"
 	"github.com/supergiant/control/pkg/clouds"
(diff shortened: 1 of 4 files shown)


pkg/workflows/steps/drain/drain_node.go
```diff
@@ -84,7 +85,7 @@ func (s *Step) Run(ctx context.Context, out io.Writer, config *steps.Config) err
 	err = steps.RunTemplate(ctx, s.script, r, out, config.DrainConfig)
 
 	if err != nil {
-		return errors.Wrap(err, "drain step")
+		logrus.Errorf("drain step has failed with %v", err)
```
saheienko: Is this step used for node deletion? If so we could delete an instance with some workloads.

stgleb: @saheienko if node fails on provisioning before register itself to kube-apiserver this step will always fail, because I get node name by ip address from kubectl get no output. So this leads to situation that node that has failed to provision can not be deleted from control. This is why I've made this step optional.

saheienko: Would it be better to retrieve node name with a kubernetes client here and check the `not found` error?

PR #1460, opened 2019-05-17
Intergrate with capacity

stgleb:
Fix master duplication
Fix getting metric for capacity made nodes


eldridgeh:
waiting on the chart repo to be reindexed and will complete testing when done. We also need to cut a release as soon as this is passed and merged in.

PR #1465, opened 2019-05-30
List releases of operational only clusters

stgleb:


pkg/kube/handler.go
```diff
@@ -913,8 +913,22 @@ func (h *Handler) getRelease(w http.ResponseWriter, r *http.Request) {
...
+		if sgerrors.IsNotFound(err) {
+			message.SendNotFound(w, kubeID, err)
+			return
+		}
+		message.SendUnknownError(w, err)
+		return
+	}
+
+	if k.State != model.StateOperational {
+		return
```
saheienko: Could you add some information to the response? It returns Status OK but a cluster isn't ready.

stgleb: Done

PR #1472, opened 2019-06-21
Add upgrade feature for k8s cluster

stgleb:
Add API endpoint

Request

`PATCH /kubes/<kube-id>`

Response

```
{
   "node-name-1":"task-id",
   "node-name-2":"task-id"
}
```

map that maps node-name to task id. 

Note: task for upgrading a node has the same id as task for provision a node. So there is no need to change anything on the UI.

Need to add some picture like spinning wheen for the cluster in provisioning state.


templates/upgrade.go
```diff
@@ -0,0 +1,27 @@
...
+const upgradeTpl = `
+sudo apt update
+sudo apt-cache policy kubeadm
+
+sudo apt-mark unhold kubeadm && \
+sudo apt-get update && sudo apt-get install -y kubeadm={{ .K8SVersion }}-00 && \
+sudo apt-mark hold kubeadm
+
+
+{{ if .IsBootstrap }}
```
saheienko: What does this mean (`IsBootstrap`) for the upgrade process?

stgleb: First you run upgrade plan that creates upgrade plan for bootstrap controlplane node and just run upgrade node on the rest of the nodes

pkg/kube/util_test.go
```diff
@@ -185,3 +185,54 @@ func TestKubeFromKubeConfig(t *testing.T) {
...
+		{
+			"empty current version",
+			"",
+			[]string{"1.11.5", "1.12.7", "1.13.7", "1.14.3"},
+			"",
+		},
+		{
+			"malformed current version",
+			"1.1",
+			[]string{"1.11.5", "1.12.7", "1.13.7", "1.14.3"},
```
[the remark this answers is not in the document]
saheienko: We could use semver library for this case: https://github.com/coreos/go-semver

PR #1487, opened 2019-07-09
remove sensitive data from logs

stgleb:


pkg/provisioner/provisioner.go
```diff
@@ -841,15 +841,15 @@ func (tp *TaskProvisioner) monitorClusterState(ctx context.Context,
...
+			logrus.Debugf("update kube %s with config", clusterID)
 			k, err := tp.kubeService.Get(ctx, clusterID)
 
 			if err != nil {
 				logrus.Errorf("cluster monitor: update kube state caused %v", err)
 				continue
 			}
 
-			logrus.Debugf("update kube %s with config %v", k.ID, config)
+			logrus.Debugf("update kube %s with config", k.ID)
```
saheienko: This comment doesn't provide any information. Should we keep it?

pkg/provisioner/provisioner.go
```diff
@@ -841,15 +841,15 @@ func (tp *TaskProvisioner) monitorClusterState(ctx context.Context,
 				continue
 			}
 		case config := <-configChan:
-			logrus.Debugf("update kube %s with config %+v", clusterID, config)
+			logrus.Debugf("update kube %s with config", clusterID)
```
saheienko: This comment doesn't provide any information. Should we keep it?

pkg/provisioner/provisioner.go
stgleb: Well it notifies about updating kube state, this can be useful for debug.

pkg/provisioner/provisioner.go
stgleb: the same goes here

PR #1488, opened 2019-07-09
make volume size configurable for azure

stgleb:


pkg/workflows/steps/azure/create_vm.go
```diff
@@ -171,7 +179,7 @@ func (s *CreateVMStep) setupVM(ctx context.Context, config *steps.Config, vmName
 						CreateOption: compute.DiskCreateOptionTypesFromImage,
 						Caching:      compute.CachingTypesReadWrite,
 						OsType:       compute.Linux,
-						DiskSizeGB:   to.Int32Ptr(30), // TODO: make it configurable
+						DiskSizeGB:   &volumeSize32, // TODO: make it configurable
```
saheienko: There is no need for the comment now.

pkg/workflows/steps/config.go
```diff
@@ -363,6 +364,8 @@ func NewConfig(clusterName, cloudAccountName string, profile profile.Profile) (*
 		AzureConfig: AzureConfig{
 			Location: profile.Region,
 			VNetCIDR: profile.CloudSpecificSettings[clouds.AzureVNetCIDR],
+			// TODO(stgleb): this should be passed from the UI
+			VolumeSize: "30",
```
saheienko: Could control check provided volume size and set the default value if it's empty?

stgleb: Provided from where? Right now we don't have any provision of volume size.

PR #1489, opened 2019-07-09
Add supergiant user

stgleb:


pkg/workflows/steps/config.go
```diff
@@ -324,12 +326,23 @@ type ConfigMap struct {
 
 // NewConfig builds instance of config for provisioning
 func NewConfig(clusterName, cloudAccountName string, profile profile.Profile) (*Config, error) {
+	var  user string
+
+	if profile.Provider == clouds.AWS {
+		//on aws default user name on ubuntu images are not root but ubuntu
+		//https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/AccessingInstancesLinux.html
+		// TODO: this should be set by provisioner
+		user = "ubuntu"
```
saheienko: This username is used only for creating the supergiant one (which is using further). Could we move all such user-related operations to the `ssh` step?

stgleb: Well, currently problem is - I can't create user with passwordless sudo in ssh session. I always require a password. So I create supergiant user, but I use ubuntu for the rest of operations.

PR #1493, opened 2019-07-12
Discovery helm version

stgleb:
Discovery helm version when import cluster

### DIFF
--- a/pkg/kube/util.go
+++ b/pkg/kube/util.go
@@ -254,5 +254,5 @@ func discoverHelmVersion(kubeConfig *clientcmddapi.Config) (string, error) {
 		}
 	}
 
-	return "", sgerrors.ErrNotFound
+	return "", nil
 }
--- a/pkg/kube/util.go
+++ b/pkg/kube/util.go
@@ -3,6 +3,8 @@ package kube
 import (
 	"context"
 	"fmt"
+	"strings"
+
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/sirupsen/logrus"
@@ -16,7 +18,6 @@ import (
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
-	"strings"
 
 	"k8s.io/client-go/kubernetes"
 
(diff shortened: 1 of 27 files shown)


pkg/kube/util.go
```diff
@@ -164,11 +172,87 @@ func findNextMinorVersion(current string, versions []string) string {
...
+			return versions[i+1]
 		}
 	}
 
 	return ""
-}
\ No newline at end of file
+}
+
+func discoverK8SVersion(kubeConfig *clientcmddapi.Config) (string, error) {
```
saheienko: There is another PR for the k8s version: https://github.com/supergiant/control/pull/1494

pkg/kube/util.go
```diff
@@ -164,11 +172,87 @@ func findNextMinorVersion(current string, versions []string) string {
...
+	deploymentList, err := clientSet.AppsV1().Deployments("kube-system").List(v1.ListOptions{})
+
+	if err != nil {
+		return "", errors.Wrapf(err, "list deployments")
+	}
+
+	for _, deployment := range deploymentList.Items {
+		if strings.Contains(deployment.Name, "tiller") {
+			for _, container := range deployment.Spec.Template.Spec.Containers {
+				slice := strings.Split(container.Image, ":")
```
saheienko: Helm client has `GetVersion` function. Please, you this function to retrieve a version.
https://github.com/helm/helm/blob/2e55dbe1fdb5fdb96b75ff144a339489417b146b/pkg/helm/interface.go#L36

templates/bootstrap_token.go
```diff
@@ -4,6 +4,9 @@ const bootstrapTokenTpl = `
 {{ if .IsBootstrap }}
 sudo kubeadm token create {{ .Token }} --ttl 0
 # Bind uploaded certs secret to bootstrap token
-sudo kubeadm init phase upload-certs --experimental-upload-certs --certificate-key {{ .CertificateKey }}
+
+{{ if not .IsImport }}
+	sudo kubeadm init phase upload-certs --upload-certs --certificate-key {{ .CertificateKey }}
```
saheienko: This change exists in the https://github.com/supergiant/control/pull/1494 PR too. Please, open another PR for this fix.

pkg/kube/util.go
stgleb: yeah, i've created this branch from branch with PR #1494 , once it is merged this PR will shrink

templates/bootstrap_token.go
stgleb: answered above

pkg/kube/util.go
stgleb: The same thing,  I can't create helm client without having a kube, but kube doesn't exists at that time.

stgleb: Actually we can do it in future, but this will be significant change for not only this handler and is out of scope of that task, I'll leave a todo for it.

saheienko: Our helm [client](https://github.com/supergiant/control/blob/master/pkg/sghelm/proxy/proxy.go#L56) doesn't require it.

PR #1494, opened 2019-07-12
Discovery k8s version

stgleb:
Discovery kubernetes version when import a cluster

### DIFF
--- a/pkg/kube/util.go
+++ b/pkg/kube/util.go
@@ -10,6 +10,11 @@ import (
 	"github.com/supergiant/control/pkg/util"
 	"github.com/supergiant/control/pkg/workflows/steps"
 	"github.com/supergiant/control/pkg/workflows/steps/amazon"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
+	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/kubernetes/scheme"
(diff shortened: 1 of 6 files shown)


pkg/kube/util.go
```diff
@@ -171,4 +176,36 @@ func findNextMinorVersion(current string, versions []string) string {
...
+	if err != nil {
+		return "", errors.Wrapf(err, "create rest config")
+	}
+
+	restConf.NegotiatedSerializer = serializer.DirectCodecFactory{CodecFactory: scheme.Codecs}
+	if len(restConf.UserAgent) == 0 {
+		restConf.UserAgent = rest.DefaultKubernetesUserAgent()
+	}
+
+	discoveryClient, err := discovery.NewDiscoveryClientForConfig(restConf)
```
saheienko: You could expose and reuse this function: https://github.com/supergiant/control/blob/master/pkg/kubeconfig/kubeconfig.go#L53

stgleb: That function is called before we have a kube https://github.com/supergiant/control/pull/1494/files/257717b62c4346500a197ff9b4de8838391be23c#diff-584d8678afd9d291396a1ca3af488e6eR1338
, so there is no way to employ that function.

PR #1503, opened 2019-08-05
Fix issue with retry

stgleb:


cmd/controlplane/main.go
```diff
@@ -50,8 +50,8 @@ func main() {
 		StorageURI:    *storageURI,
 		TemplatesDir:  *templatesDir,
 		LogDir:        *logDir,
-		ReadTimeout:   time.Second * 20,
-		WriteTimeout:  time.Second * 10,
+		ReadTimeout:   time.Second * 60,
```
taaraora: helm cli has a default timeout for individual operations equal to 300s.
Let's just have the same timeout?

stgleb: Done

